PR #83, opened 2021-03-02
Time phases individually

basile-henry:
This should make it a bit easier when debugging to see which phases take a long time.

(I'm not sure `Δ` is the best description for what the phase time is ^^)


zimbatm (approved):
nice!

PR #87, opened 2021-03-02
Code suggestion: Mutate the eval cache instead of cloning it

ldesgoui:
Hey, as per my comments on twitter, this would be one way to reduce the number of allocations in `eval_cache`, I'll sprinkle some comments on the diff to help making the changes understandable, feel free to ask any questions.

### DIFF
--- a/src/eval_cache.rs
+++ b/src/eval_cache.rs
@@ -9,10 +9,10 @@ use anyhow::Result;
 use log::{debug, error, warn};
 use serde::{Deserialize, Serialize};
 use sha1::{Digest, Sha1};
+use std::collections::BTreeMap;
 use std::fs::{read_to_string, File};
 use std::io::Write;
 use std::path::{Path, PathBuf};
-use std::{borrow::BorrowMut, collections::BTreeMap};
 
 /// Metadata about the formatter
 #[derive(Debug, Deserialize, Serialize, PartialEq, Eq, Clone)]
@@ -106,21 +106,19 @@ impl CacheManifest {
 
     /// Checks and inserts the formatter info into the cache.
     /// If the formatter info has changed, invalidate all the old paths.
-    #[must_use]
-    pub fn update_formatters(self, formatters: BTreeMap<FormatterName, Formatter>) -> Self {
-        let mut new_formatters = BTreeMap::new();
-        let mut new_paths = self.matches.clone();
+    pub fn update_formatters(&mut self, formatters: BTreeMap<FormatterName, Formatter>) {
+        let mut old_formatters = std::mem::replace(&mut self.formatters, BTreeMap::new());
         for (name, fmt) in formatters {
-            match load_formatter_info(&fmt) {
+            match load_formatter_info(fmt) {
                 Ok(new_fmt_info) => {
-                    if let Some(old_fmt_info) = self.formatters.get(&name) {
+                    if let Some(old_fmt_info) = old_formatters.remove(&name) {
                         // Invalidate the old paths if the formatter config has changed.
-                        if old_fmt_info != &new_fmt_info {
-                            new_paths.remove(&name);
+                        if old_fmt_info != new_fmt_info {
+                            self.matches.remove(&name);
                         }
                     }
                     // Record the new formatter info
-                    new_formatters.insert(name, new_fmt_info);
+                    self.formatters.insert(name, new_fmt_info);
                 }
                 Err(err) => {
                     // TODO: This probably means that there is a deeper issue with the formatter and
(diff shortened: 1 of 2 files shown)


src/eval_cache.rs
```diff
@@ -133,23 +131,15 @@ impl CacheManifest {
...
+        for name in old_formatters.keys() {
+            self.matches.remove(&name);
         }
     }
 
     /// Returns a new map with all the paths that haven't changed
     #[must_use]
     pub fn filter_matches(
-        self,
+        &self,
```
ldesgoui, line 142: That's all it took to get rid of a significant clone, because we're only reading from `self` and we don't need to "consume" it, we can just use a reference

src/eval_cache.rs
```diff
@@ -108,21 +108,19 @@ impl CacheManifest {
 
     /// Checks and inserts the formatter info into the cache.
     /// If the formatter info has changed, invalidate all the old paths.
-    #[must_use]
-    pub fn update_formatters(self, formatters: BTreeMap<FormatterName, Formatter>) -> Self {
-        let mut new_formatters = BTreeMap::new();
-        let mut new_paths = self.matches.clone();
+    pub fn update_formatters(&mut self, formatters: BTreeMap<FormatterName, Formatter>) {
+        let mut old_formatters = std::mem::replace(&mut self.formatters, BTreeMap::new());
```
ldesgoui, line 112: This is basically a swap, we're setting `self.formatters` to `BTreeMap::new()` and putting the old `self.formatters` in the variable on the left.

src/eval_cache.rs
```diff
@@ -108,21 +108,19 @@ impl CacheManifest {
...
         for (name, fmt) in formatters {
-            match load_formatter_info(&fmt) {
+            match load_formatter_info(fmt) {
```
ldesgoui, line 114: Because we already have a whole cloned "copy" of `formatters` (and therefore `fmt`) in this instance, we might as well just consume them to prevent some clones below

src/eval_cache.rs
```diff
@@ -108,21 +108,19 @@ impl CacheManifest {
...
                 Ok(new_fmt_info) => {
-                    if let Some(old_fmt_info) = self.formatters.get(&name) {
+                    if let Some(old_fmt_info) = old_formatters.remove(&name) {
                         // Invalidate the old paths if the formatter config has changed.
-                        if old_fmt_info != &new_fmt_info {
-                            new_paths.remove(&name);
+                        if old_fmt_info != new_fmt_info {
```
ldesgoui, line 118: This is a small logic change; we're draining values off `old_formatters` so that when the iteration is done, what remains in it is the set difference compared to `formatters`
We remove `&` because `old_fmt_info` went from `&FormatterInfo` to `FormatterInfo`. They now have the same type, so we can compare directly.

src/eval_cache.rs
```diff
@@ -133,23 +131,15 @@ impl CacheManifest {
...
-                new_paths.remove(&name);
-            }
-        }
-
-        // Replace with the new info
-        Self {
-            formatters: new_formatters,
-            matches: new_paths,
+        for name in old_formatters.keys() {
+            self.matches.remove(&name);
```
ldesgoui, line 135: And here we can just remove using the difference

src/eval_cache.rs
```diff
@@ -177,41 +167,26 @@ impl CacheManifest {
...
-            let mut def = BTreeMap::new();
-            let merged_path_infos = new_matches_cmp
-                .get_mut(&name)
-                .unwrap_or_else(|| def.borrow_mut());
-            for (path, mtime) in path_infos {
-                merged_path_infos.insert(path.clone(), mtime);
+            if let Some(old_path_infos) = self.matches.get_mut(&name) {
+                old_path_infos.extend(path_infos);
+            } else {
+                self.matches.insert(name, path_infos);
```
ldesgoui, line 175: "if there's a value attached to the name, merge it with the new `path_infos`, otherwise set that name to the new `path_infos`"

src/eval_cache.rs
```diff
@@ -177,41 +167,26 @@ impl CacheManifest {
...
+fn load_formatter_info(fmt: Formatter) -> Result<FormatterInfo> {
+    let command = fmt.command;
     let command_mtime = get_path_mtime(&command)?;
     // Resolve symlinks and everything
     let command_resolved = std::fs::canonicalize(command.clone())?;
     let command_resolved_mtime = get_path_mtime(&command_resolved)?;
-    let options = fmt.options.clone();
-    let work_dir = fmt.work_dir.clone();
+    let options = fmt.options;
+    let work_dir = fmt.work_dir;
```
ldesgoui, line 189: Here, we can just move the values off the `Formatter` struct because it's about to drop, and reuse them for other stuff

Rizary:
Thank you for the suggestion!!

Rizary (approved):

zimbatm:
thanks! I learned a lot from this PR

PR #98, opened 2021-03-31
treat exit status code > 0 as an error

Rizary:
To try this PR:
1. edit `examples/rust/src/main.rs` by removing 1 syntax.
2. run `cargo run --`

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -183,7 +183,14 @@ pub fn run_treefmt(
 
                 match formatter.clone().fmt(&paths) {
                     // FIXME: do we care about the output?
-                    Ok(_) => {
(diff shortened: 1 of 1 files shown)


src/engine.rs
```diff
@@ -183,7 +183,14 @@ pub fn run_treefmt(
 
                 match formatter.clone().fmt(&paths) {
                     // FIXME: do we care about the output?
-                    Ok(_) => {
+                    Ok(out) => {
```
zimbatm, line 186: it's probably best to use `out.status.success()` since it's available on all platforms. Then special-case the error message whenever status.code() exists or not.

PR #101, opened 2021-04-08
Handling output formatter

Rizary:


src/formatter.rs
```diff
@@ -98,7 +98,11 @@ impl Formatter {
         // Append all of the file paths to format.
         cmd_arg.args(paths);
         // And run
-        Ok(cmd_arg.output()?)
+        match cmd_arg.output() {
+            Ok(out) => Ok(out),
+            Err(err) => Err(anyhow!("formatting error due to {}", err))
+        }
+        // Ok(cmd_arg.output()?)
```
zimbatm, line 105: Can you change the

src/engine.rs
```diff
@@ -177,13 +177,31 @@ pub fn run_treefmt(
...
+                                    ));
+                                }
+                                None => {
+                                    return Err(anyhow!(
+                                        "{}'s formatter failed: unknown formatter error",
+                                        &formatter
+                                    ));
+                                }
+                            }
+                        }
```
zimbatm, line 203: Instead of exposing the Output, this should be handled in the Formatter fmt function. Then change the return signature of the function to `Result<()>`.

src/engine.rs
```diff
@@ -199,18 +217,17 @@ pub fn run_treefmt(
...
-                        error!("{} failed: {}", &formatter, err);
                         // Assume the paths were not formatted
-                        (formatter_name.clone(), path_mtime.clone())
+                        return Err(anyhow!("{} failed: {}", &formatter, err));
                     }
                 }
             }
         })
-        .collect::<BTreeMap<FormatterName, BTreeMap<PathBuf, Mtime>>>();
+        .collect::<Result<BTreeMap<FormatterName, BTreeMap<PathBuf, Mtime>>, _>>()?;
```
zimbatm, line 230: Is there no code later on that takes that into account?

If one formatter fails, the exit status of `treefmt` should also reflect that.

PR #121, opened 2021-07-16
Support global excludes

### LINKED ISSUE #119: global excludes
I have third_party code in my repo, and I don't want to reformat all this to make treefmt happy.
I also don't want to list that as an exclude for each formatter I defined in `treefmt.toml`

A global exclude pattern list, that is added to the excludes of all formatters.

Defining that exclude for all formatters, but it gets very verbose quite quickly.

basile-henry:
Closes #119

Adds an extra field to the config to support excludes that apply to all formatters

### DIFF
--- a/docs/treefmt-configuration.md
+++ b/docs/treefmt-configuration.md
@@ -22,3 +22,12 @@ This section describes the integration between a single formatter and
 
 - `excludes`: A list of glob patterns to deny. If any of these patterns match,
   the file will be excluded.
+
+## Global config
+
+This section describes the config that applies to every formatters.
+In the top level of `treefmt.toml`:
+
+- `excludes`: A list of glob patterns to deny. If any of these patterns match,
+  the file will be excluded. This list is appended to individual formatters'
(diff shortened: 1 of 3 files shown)


docs/treefmt-configuration.md
```diff
@@ -22,3 +22,12 @@ This section describes the integration between a single formatter and
 
 - `excludes`: A list of glob patterns to deny. If any of these patterns match,
   the file will be excluded.
+
+## Global config
+
+This section describes the config that applies to every formatters.
+In the top level of `treefmt.toml`:
```
zimbatm, line 29: Do you mind moving these in a section? I noticed that top-level attributes are a bit awkward to use because the ordering matters. The section can be named `[global]` or something else if you have a preference.

zimbatm:
nice! thanks for looking into this

docs/treefmt-configuration.md
basile-henry: Good idea :+1:

PR #131, opened 2021-10-07
allow config files to be independent from worktree location

tazjin:
This is useful in cases where a config file is built by Nix (e.g. to ensure the availability of all formatters by templating their paths), as would be preferred for the TVL monorepo.


tazjin:
> Can you also expand the path if the config-file is passed?

Done.

PR #134, opened 2021-11-11
Trade in naersk for a newer version of nixpkgs

blaggacao:
- Naersk has the annoying property to pin nixpkgs as a dep
- Nixpkgs mostly is a _core_ dependency that it is worth auditing duely
  (e.g. to avoid cache mismatches and other spourious errors)
- Every usage of naersk adds a another "spurious" nixpkgs dep somewhere
  to the flake lock. That makes auditing less straight forward.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -13,13 +13,15 @@ let
     pkgs = nixpkgs;
   };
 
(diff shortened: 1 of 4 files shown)


zimbatm:
macOS is broken in nixpkgs I think

PR #161, opened 2022-05-03
nix: move ci dependencies to the devShell

### LINKED ISSUE #160: The default nix flake installable is heavy

Running `nix run github:numtide/treefmt` pulls about 5GB of dependencies.
It seems like the linters are somehow downloaded even when they are not used.

The default app should be lightweight and downloading the linters tools should be optional.

TristanCacqueray:
Fixes #160

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -17,8 +17,23 @@ let
 
     src = nixpkgs.lib.cleanSource ./.;
 
-    # Those are being used in tests
-    nativeBuildInputs = with nixpkgs; [
+    buildInputs = with nixpkgs; lib.optionals stdenv.isDarwin [ darwin.apple_sdk.frameworks.Security libiconv ];
+
+    doCheck = true;
+
+    cargoLock.lockFile = ./Cargo.lock;
+
+    meta.description = "one CLI to format the code tree";
+  };
+
+  # Add all the dependencies of treefmt, plus more build tools
+  devShell = treefmt.overrideAttrs (prev: {
+    shellHook = ''
+      # Put the treefmt binary on the PATH when it's built
+      export PATH=$PWD/target/debug:$PATH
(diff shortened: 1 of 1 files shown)


zimbatm:
```
 = note: ld: library not found for -liconv
          clang-7: error: linker command failed with exit code 1 (use -v to see invocation)
```
I'm not sure why this is failing on macOS

TristanCacqueray:
@zimbatm shouldn't the override appends to the treefmt buildInputs? e.g. to keep that definition: https://github.com/numtide/treefmt/blob/e7918e97ce19a5fbe91c0eb6bf336ed39f92f99d/default.nix#L49

zimbatm:
doh. that's probably it :)

default.nix
```diff
@@ -17,8 +17,23 @@ let
...
+  };
+
+  # Add all the dependencies of treefmt, plus more build tools
+  devShell = treefmt.overrideAttrs (prev: {
+    shellHook = ''
+      # Put the treefmt binary on the PATH when it's built
+      export PATH=$PWD/target/debug:$PATH
+    '';
+
+    buildInputs = prev.buildInputs ++ (with nixpkgs; [
```
TristanCacqueray, line 36: Oops, I think that should be:

```suggestion
    nativeBuildInputs = (with nixpkgs; [
```

zimbatm:
@TristanCacqueray nativeBuildInputs probably also needs to be concatenated with prev.nativeBuildInputs

TristanCacqueray:
@zimbatm oh, ok I've squashed the change. We'll get through this eventually :)

zimbatm:
success!

TristanCacqueray:
nice, thanks!

Issue #78, opened 2021-02-27
Add --pre-commit mode


zimbatm:
Some developers like to invoke the formatters just before committing the code.

When invoked, it should only format the code that has been staged for the commit and not touch the rest.

This can be done like this: https://stackoverflow.com/a/26911078
And also probably check for unresolved merge commits before and abort if that's the case.

ocharles:
I've ended up using `pre-commit` with this:

```yaml
repos:
  - repo: local
    hooks:
      - id: treefmt
        name: treefmt
        entry: treefmt
        language: system
        always_run: true
```

Seems to work well!

loicreynier:
A `.pre-commit-hooks.yaml` file could also be added to this repository such as 

```yaml
- id: treefmt
  name: treefmt
  description: Format code with treefmt
  entry: treefmt
  language: rust
  minimum_pre_commit_version: 1.18.1
```

so that `pre-commit` could download and install (with Cargo and from crates) `treefmt` with

```yaml
repos:
  - repo: https://github.com/numtide/treefmt
    rev: master
    hooks:
      - id: treefmt
```

zimbatm:
The pre-commit framework is a bit heavy for my taste so I wrote a bash script to replace it (see #166). I think both approaches are valid, depending on if you use nix to pull the dependencies or not. If you don't then it's nice to have all the machinery that's provided by the pre-commit python code.

nikolay:
@zimbatm We do use it though, and I've personally been waiting for [Lefthook](https://github.com/evilmartians/lefthook) to take over, but it hasn't happened yet, and pre-commit is the king still. Adding a simple YAML file here in the repo would solve elegantly the problem for those of us having to use pre-commit for the time being.

brianmcgee:
This is now being tracked here https://github.com/numtide/treefmt/issues/311

PR #166, opened 2022-05-12
add pre-commit contrib script

zimbatm:
This is working towards #78 . In the end that script would be shipped
with the treefmt binary, but for now let's just test it out.


zimbatm:
One known issue I found out is that if a file has a mix of staged and unstaged changes, the unstaged changes get lost. I'm not sure yet if the issue is with `git stash push` or `git stash pop`

Issue #167, opened 2022-05-12
treefmt cli: allow explicitly specifying a specific formatter


flokli:
Sometimes `treefmt` with a bunch of formatters is a bit slow. In some cases, I'd like to explicitly narrow down the list of formatters to run.

It'd be nice if I could only run a subset of formatters, by invoking `treefmt -f nix,terraform` (or similar syntax)

zimbatm:
I like this. :+1:

brianmcgee:
@zimbatm should this issue be closed as it looks like it was already merged?

zimbatm:
thanks

PR #168, opened 2022-05-18
cli: add --formatters option

jfroche:
Select formatters name to apply. Defaults to all formatters

refs #167


clippy.toml
```diff
@@ -1 +1 @@
-too-many-arguments-threshold = 8
+too-many-arguments-threshold = 9
```
zimbatm, line 1: good call :-D

zimbatm:
thanks!

Issue #165, opened 2022-05-11
Using treefmt in flake projects


srid:
I'd like to minimize the Nix stub required to use treefmt in my projects. Presently, there is a ~40 line nix that provides me `nix run .#format` (runs treefmt in project tree) as well a flake check that is built in CI. Could these be upstreamed to treefmt's flake in some way such that I can use it as a flake input, and directly use those app and check without replicating it in every project? Or is there a better approach to this?

Here's the ~40 line Nix:

```nix
{ inputs, system, pkgs }:

# Uses https://github.com/numtide/treefmt to provide auto-formating for multiple
# languages. Configure the behaviour in treefmt.toml at project root.
rec {
  dependencies = [
    pkgs.haskellPackages.cabal-fmt
    pkgs.haskellPackages.fourmolu
    pkgs.treefmt
  ];

  apps = {
    format = {
      type = "app";
      program = pkgs.writeShellApplication
        {
          name = "format";
          runtimeInputs = dependencies;
          text = "treefmt";
        } + "/bin/format";
    };
  };

  checks = {
    # Checks that the project tree is *already* formatted.
    format = pkgs.runCommandLocal "format-check"
      {
        buildInputs = dependencies;
      } ''
      set -e
      # treefmt maintains a cache at ~/.cache; so use $TMP as home.
      # treefmt 0.4 has a --no-cache option, but our nixpkgs has only 0.3
      export HOME=$TMP
      treefmt -vvv --clear-cache --fail-on-change -C ${inputs.self}
      touch $out
    '';
  };
}
```

I imagine we want to parameterize this over 'dependencies'.

srid:
@zimbatm That's a great idea but it is tangential to this issue? Because, while it would get rid of `dependencies`, I'd still have to hand-write `apps.format` and `checks.format`.

zimbatm:
Yup. That shape depends on how the systems get abstracted away.

Nix 2.8 also has a `formatters.<system> = <installable>` interface so you could run `nix fmt` that then invokes `treefmt`, that then invokes the rest :p

MatthewCroughan:
I would prefer to see `nix fmt` win, seems like the right approach.

zimbatm:
Technically, `nix fmt` leaves the filesystem traversal up to the tool. But I don't know if that is an invariant or not. (and also if it's a good idea).

zimbatm:
[...]

PR #169, opened 2022-05-20
add treefmt.withConfig

zimbatm:
This is useful if you want to configure treefmt with nix, and precisely
pass all the commands from nixpkgs.

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,7 @@
+#!/usr/bin/env bash
+# Environment variables for the project. See https://direnv.net/
+if nix flake metadata &>/dev/null; then
+  use flake
+else
+  use nix
+fi
--- a/default.nix
+++ b/default.nix
@@ -13,6 +13,18 @@ let
 
   cargoToml = with builtins; (fromTOML (readFile ./Cargo.toml));
 
+  # Use the Nix module system to validate the treefmt config file format.
+  evalModule = config:
+    lib.evalModules {
+      modules = [
+        {
+          _module.args = { inherit nixpkgs lib treefmt; };
+        }
+        ./module-options.nix
+        config
+      ];
+    };
+
   # What is used when invoking `nix run github:numtide/treefmt`
   treefmt = rustPackages.rustPlatform.buildRustPackage {
     inherit (cargoToml.package) name version;
@@ -34,6 +46,12 @@ let
     cargoLock.lockFile = ./Cargo.lock;
 
     meta.description = "one CLI to format the code tree";
+
+    passthru.withConfig = config:
+      let
+        mod = evalModule config;
+      in
+      mod.config.build.wrapper;
   };
 
   # Add all the dependencies of treefmt, plus more build tools
@@ -68,7 +86,13 @@ let
   });
 in
 {
-  inherit treefmt devShell;
+  inherit treefmt devShell evalModule;
+
+  # module that generates and wraps the treefmt config with Nix
+  module = ./module-options.nix;
+
+  # reduce a bit of repetition
+  inherit (treefmt.passthru) withConfig;
 
   # A collection of packages for the project
   docs = nixpkgs.callPackage ./docs { };
--- a/flake.nix
+++ b/flake.nix
@@ -18,9 +18,22 @@
           };
         in
         {
-          inherit packages;
+          # This contains a mix of packages, modules, ...
+          legacyPackages = packages;
 
           devShells.default = packages.devShell;
+
+          # In Nix 2.8 you can run `nix fmt` to format this whole repo.
+          #
+          # Because we load the treefmt.toml and don't define links to the
+          # packages in Nix, the formatter has to run inside of `nix develop`
+          # to have the various tools on the PATH.
+          #
+          # It also assumes that the project root has a flake.nix (override this by setting `projectRootFile`).
+          formatter = packages.treefmt.withConfig {
+            settings = nixpkgs.lib.importTOML ./treefmt.toml;
+            projectRootFile = "flake.nix";
+          };
         };
     };
 }
--- a/flake.lock
+++ b/flake.lock
@@ -7,11 +7,11 @@
         ]
       },
       "locked": {
-        "lastModified": 1661009076,
-        "narHash": "sha256-phAE40gctVygRq3G3B6LhvD7u2qdQT21xsz8DdRDYFo=",
+        "lastModified": 1663686189,
+        "narHash": "sha256-cgWGo0SLM4RX0DqodeFHsU0PRGlmz6AOSF1lmX21bvg=",
         "owner": "hercules-ci",
         "repo": "flake-parts",
-        "rev": "850d8a76026127ef02f040fb0dcfdb8b749dd9d9",
+        "rev": "f17e9dba098c0a7bca10029bec5384fd6425c157",
         "type": "github"
       },
       "original": {
@@ -22,11 +22,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1661239106,
-        "narHash": "sha256-C5OCLnrv2c4CHs9DMEtYKkjJmGL7ySAZ1PqPkHBonxQ=",
+        "lastModified": 1663669617,
+        "narHash": "sha256-yUrzkRDc6P3hsI5TdQ5+q8gnyjadNvaM3MMEsEVS8qk=",
         "owner": "NixOS",
         "repo": "nixpkgs",
-        "rev": "963d27a0767422be9b8686a8493dcade6acee992",
+        "rev": "41ac0bd371618db6dd67fd952cc5b3d6a9955a15",
         "type": "github"
       },
       "original": {
--- a/module-options.nix
+++ b/module-options.nix
@@ -1,0 +1,113 @@
+{ lib, nixpkgs, treefmt, ... }:
+let
+  # A new kind of option type that calls lib.getExe on derivations
+  exeType = lib.mkOptionType {
+    name = "exe";
+    description = "Path to executable";
+    check = (x: lib.isString x || builtins.isPath x || lib.isDerivation x);
+    merge = loc: defs:
+      let res = lib.mergeOneOption loc defs; in
+      if lib.isString res || builtins.isPath res then
+        "${res}"
+      else
+        lib.getExe res;
+  };
+
+  # The schema of the treefmt.toml data structure.
+  configSchema = with lib; {
+    excludes = mkOption {
+      description = "A global list of paths to exclude. Supports glob.";
+      type = types.listOf types.str;
+      default = [ ];
+      example = [ "./node_modules/**" ];
+    };
+
+    formatter = mkOption {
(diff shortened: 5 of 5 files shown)


srid:
cf https://github.com/numtide/treefmt/issues/165#issuecomment-1124036328

.envrc
```diff
@@ -0,0 +1,2 @@
+export PRJ_ROOT=$PWD
```
srid, line 1: Interesting trick! Does `direnv` also use project root as `PWD` even if you open the terminal directly in one of its sub directories?

srid:
@zimbatm What do you think of exposing a `flakeModule` (a la https://github.com/cachix/pre-commit-hooks.nix/pull/130) so that projects that already use `flake-parts` can use this directly, thus resolving #165?

.envrc
zimbatm: Yes, this is necessary so that relative paths in the `.envrc` have the same result, even if the user jumps into a sub-folder directly.

zimbatm:
Sounds good, the flake module probably can call treefmt.withConfig with its resulting schema at the end.

default.nix
```diff
@@ -24,13 +28,28 @@ let
     cargoLock.lockFile = ./Cargo.lock;
 
     meta.description = "one CLI to format the code tree";
+
+    passthru.withConfig = { ... }@config:
```
zimbatm, line 32: TODO: use the module system and expose the settings as well.

srid:
> Sounds good, the flake module probably can call treefmt.withConfig with its resulting schema at the end.

Great; I'd be happy to test it out here:

https://github.com/srid/haskell-template/blob/6c9e2eb0870b108731df87cdcbb016c064e55041/flake.nix#L19-L25

default.nix
zimbatm: Done

zimbatm:
@srid ok this is ready if you want to give it a try

srid:
@zimbatm Trying it out on haskell-template,

```diff
diff --git a/flake.nix b/flake.nix
index 9918a57..684242e 100644
--- a/flake.nix
+++ b/flake.nix
@@ -5,22 +5,23 @@
     flake-parts.url = "github:hercules-ci/flake-parts";
     flake-parts.inputs.nixpkgs.follows = "nixpkgs";
     haskell-flake.url = "github:srid/haskell-flake";
+    treefmt.url = "github:numtide/treefmt/withConfig";
+    treefmt.flake = false;
   };
 
-  outputs = { self, nixpkgs, flake-parts, haskell-flake, ... }:
+  outputs = inputs@{ self, nixpkgs, flake-parts, haskell-flake, treefmt, ... }:
     flake-parts.lib.mkFlake { inherit self; } {
       systems = nixpkgs.lib.systems.flakeExposed;
       imports = [
         haskell-flake.flakeModule
       ];
-      perSystem = { self', pkgs, ... }: {
+      perSystem = { self', system, pkgs, ... }: {
         haskellProjects.default = {
           root = ./.;
           buildTools = hp: {
-            # TODO: Use https://github.com/numtide/treefmt/pull/169
             inherit (pkgs)
-              treefmt
               nixpkgs-fmt;
+            treefmt = (import treefmt { inherit system; nixpkgs = pkgs; }).treefmt.withConfig (pkgs.lib.importTOML ./treefmt.toml);
             inherit (hp)
               cabal-fmt
               fourmolu;
```

I see:

```
srid on now haskell-template on  master [!] via λ 9.0.2 via ❄️  impure (ghc-shell-for-haskell-template-0.1.0.0) 
❯ more $(which treefmt)
#!/nix/store/14lypyys4gfcl982rjddxa6jg7msqz9q-bash-5.1-p16/bin/bash
exec /nix/store/d88p6rcrpp83ir7i6lasljf871j34g25-treefmt/bin/treefmt --config-file /nix/store/vfblhmhr809df21cvcnh7mdwpqn0nv9w-treefmt.toml "$@"


srid on now haskell-template on  master [!] via λ 9.0.2 via ❄️  impure (ghc-shell-for-haskell-template-0.1.0.0) 
❯ treefmt
[ERR]: If --config-file is set, --tree-root must also be set
```

I wanted to see how `importTOML` behaved before switching to config defined entirely in Nix, but as you can see it looks like I need this PRJ_ROOT env var and pass it via `--tree-root`, which in turn would require direnv or devshell. Therefore, it seems to be that `treefmt.withConfig` as it is implemented currently wouldn't work with projects that do not use either of direnv or devshell.

zimbatm:
agreed. And since the absolute path to the project root would vary from machine to machine, I don't have a really good solution for it. Unless maybe the function also takes a command that points to the project root, like `git rev-parse --show-toplevel`

srid:
What if we had a wrapper script to determine the project root by traversing until a parent directory with a `flake.nix` in it? It does assume that the project has no sub-flakes in it.

srid:
> What if we had a wrapper script to determine the project root by traversing until a parent directory with a `flake.nix` in it? It does assume that the project has no sub-flakes in it.

We can refine this idea to avoid the sub-flake assumption as follows:

```nix
formatter = pkgs.treefmt.withConfig (nixpkgs.lib.importTOML ./treefmt.toml) { projectRootFile = "flake.nix" };
```

Basically, this tells the wrapper script to `cd ../.` until the `projectRootFile` can be located. If the project has sub-flakes, the user can set it to an alternative file, like `{ projectRootFile = "cabal.project"; }`.

@zimbatm If we can get this feature merged, it would be useful in https://github.com/srid/treefmt-flake/issues/1. Ultimately, I imagine that `treemfmt-flake` can be upstreamed to the flake.nix of this repo.

zimbatm:
yeah, that makes sense. Do you have the time to finish this PR so it takes `pkgs.treefmt.withConfig { config = ...; projectRootFile = "..."; }`, all of this gets passed to the module system, and it outputs the wrapper?

srid:
@zimbatm Let me know what you think: https://github.com/numtide/treefmt/pull/177

zimbatm:
ok, I think this is ready now. Any last words/feedback?

flake.nix
```diff
@@ -20,6 +20,14 @@
 
         packages = pkgs;
 
+        # In Nix 2.8 you can run `nix fmt` to format this whole repo. Note that you need to have loaded the
+        # `nix develop` shell before so the various formatters are available in the PATH.
```
zimbatm, line 24: that's true only if the dependencies are not specified in nix. if the config has references to packages then it should work even outside of `nix develop`

srid:
@zimbatm I'm testing this out. I had to apply this diff:

```diff
diff --git a/default.nix b/default.nix
index fc46f6e..d55fff6 100644
--- a/default.nix
+++ b/default.nix
@@ -153,7 +153,7 @@ let
       let
         mod = evalModule config;
       in
-      mod.build.wrapper;
+      mod.config.wrapper;
   };
```

But `nix fmt` still doesn't work:

image

srid:
Also `withConfig` is currently returning a set (`trace: { package = <CODE>; projectRootFile = <CODE>; }`), which forces me to use it like:

```nix
            treefmt = (inputs'.treefmt.packages.default.withConfig {
              settings = nixpkgs.lib.importTOML ./treefmt.toml;
              wrapper.projectRootFile = "flake.nix";
            }).package;
```

(Note the `.package` at the end).

Wouldn't it be simpler to be able to write it without the `.package`?  I.e, have `withConfig` directly return the package.

srid:
> ok, I think this is ready now. Any last words/feedback?

I've tested it to work after applying #178

Looks good to merge. I'd only suggest making `wrapper.projectRootFile = "flake.nix";` the default so the user doesn't have to specify it explicitly on every project using flakes (but without a sub-flake).

zimbatm:
Rebased. It looks like flake.parts doesn't support the formatter option yet. I will give it another shot next week.

default.nix
```diff
@@ -13,6 +13,120 @@ let
...
   cargoToml = with builtins; (fromTOML (readFile ./Cargo.toml));
 
+  # A new kind of option type that calls lib.getExe on derivations
+  exeType = lib.mkOptionType {
+    name = "exe";
+    description = "Path to executable";
+    check = (x: lib.isString x || lib.isPath x || lib.isDerivation x);
+    merge = loc: defs:
+      let res = lib.mergeOneOption loc defs; in
+      if lib.isString res || lib.isPath res then
```
srid, line 23: When passing a package, this is failing with:

```
error: attribute 'isPath' missing

       at /nix/store/n7v25mkag2k3g170ljfkppg5aj07hqv5-source/default.nix:20:35:

           19|     description = "Path to executable";
           20|     check = (x: lib.isString x || lib.isPath x || lib.isDerivation x);
             |                                   ^
           21|     merge = loc: defs:
```

Should it be `builtins.isPath` instead? However, `lib.isPath` should work in theory, going by [what's in nixpkgs](https://github.com/NixOS/nixpkgs/search?q=isPath&type=code).

default.nix
```diff
@@ -13,6 +13,120 @@ let
 
   cargoToml = with builtins; (fromTOML (readFile ./Cargo.toml));
 
+  # A new kind of option type that calls lib.getExe on derivations
+  exeType = lib.mkOptionType {
+    name = "exe";
+    description = "Path to executable";
+    check = (x: lib.isString x || lib.isPath x || lib.isDerivation x);
```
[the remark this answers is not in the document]
zimbatm: yes, turns out lib.isPath is not part of nixpkgs/lib

zimbatm:
Blocked on https://github.com/hercules-ci/flake-parts/pull/56

.envrc
MatthewCroughan: `DIRENV_DIR` already exists and can be queried if you use direnv, there may be no need to set `PRJ_ROOT` then.

zimbatm:
ok, I think this is ready

.envrc
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
+# Environment variables for the project. See https://direnv.net/
+if nix flake info &>/dev/null; then
```
Mic92, line 3:
```suggestion
if nix flake metadata &>/dev/null; then
```

srid:
@zimbatm Will this (`withConfig`) be upstreamed to nixpkgs at some point?

zimbatm:
I'm unsure about the policy regarding nix modules in the pkgs/ tree. Are there any precedents?

srid:
I have seen `passthru` used by certain packages, eg: https://github.com/NixOS/nixpkgs/blob/45b92369d6fafcf9e462789e98fbc735f23b5f64/nixos/modules/services/networking/stubby.nix#L31

zimbatm:
that's not quite the same, plus it's using IFD :-D

With @Mic92 we discussed moving this to another repo, so all the module systems could live separate from the code, and include pre-made samples for the various programs.

zimbatm:
Now live at https://github.com/numtide/treefmt-nix . Looking for feedback!

srid:
> Looking for feedback!

Sure: https://github.com/numtide/treefmt-nix/issues/2

PR #172, opened 2022-07-04
Add support for `.treefmt.toml`

loicreynier:
This adds support for a configuration file `.treefmt.toml` in addition to the default `treefmt.toml`. If both are found, only the `treefmt.toml` is accepted. I could also raise an error in that case.

Most of the formatters support dotfiles as configuration file (which I prefer), I figured out that the project could benefit from this option.


zimbatm:
Fine

PR #177, opened 2022-08-23
Make `treefmt.withConfig` work independent of $PRJ_ROOT

srid:
https://github.com/numtide/treefmt/pull/169#issuecomment-1219213800

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,1 @@
-export PRJ_ROOT=$PWD
 use nix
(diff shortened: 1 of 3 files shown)


srid:
@zimbatm 

> all of this gets passed to the module system, and it outputs the wrapper?

It outputs the correct wrapper, but I'm not sure what you mean by "gets passed to the module system".

zimbatm:
thanks, I'll fix it

PR #188, opened 2022-12-01
feat: hard fail on missing tools

### LINKED ISSUE #180: Hard fail on missing tools

Right now treefmt ignores commands if it cannot find the related executable. This allows developers to partially format the code tree even if they don't have all the dependencies installed.

The issue is that in CI, we want to be notified if a tool is missing.

Either add a new `--fail-on-missing-command` flag, or change the behaviour of treefmt when `--fail-on-change` is enabled.

N/A

**Additional context**

Some internal repo: https://github.com/numtide/blended/issues/288

brianmcgee:
By default if a formatter cannot be loaded you will get output like this:

```shell
[ERR]: Failed to load formatter #go due to error: cannot find binary path
[ERR]: Failed to load formatter #rust due to error: cannot find binary path
[ERR]: One or more formatters are missing
```

You can enable the current behaviour with the flag `--allow-missing-formatter` which will simply log that it will ignore the formatter.

Closes #180

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,14 +47,15 @@ USAGE:
     treefmt [FLAGS] [OPTIONS] [paths]...
 
 FLAGS:
-        --clear-cache       Clear the evaluation cache. Use in case the cache is not precise enough
-        --fail-on-change    Exit with error if any changes were made. Useful for CI
-    -h, --help              Prints help information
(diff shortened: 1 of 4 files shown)


brianmcgee:
I was operating on stale code. Will delete this PR and create a fresh one after updating

PR #200, opened 2022-12-12
wip: migration to clap

### LINKED ISSUE #194: Migrate from structop to clap
[structopt](https://docs.rs/structopt/latest/structopt/) is now in maintenance mode:

> As clap v3 is now out, and the structopt features are integrated into (almost as-is), structopt is now in maintenance mode: no new feature will be added.

It looks like a migration to clap is almost like for like, with only a few minor changes to things like the `-v` arg definition. For the most part though it looks to use the same macros and signatures.

I would recommend increasing the test coverage for the cli parsing first before making the swap in order to catch any nuanced bugs that might get introduced.

### LINKED ISSUE #195: Move post processing logic for cli into clap TypedValueParser's
Currently there is a post-processing step which ensures relative paths are expanded into absolute paths and searches for the config file if required. 

https://github.com/numtide/treefmt/blob/fc133187bd0de57192f089ee2bc6bb298fdd9913/src/command/mod.rs#L78-L111

A more natural fit for this functionality would be as custom value parsers https://docs.rs/clap/latest/clap/struct.Arg.html#method.value_parser, one for realising `PathBuf` as absolute paths and a second which optionally expands or searches for the config file. 

This is predicated upon the implementation of #194

aldoborrero:
This PR will conflate several related tasks to CLI (that I think it makes sense to bring it together):

- Closes #194
- Closes #195 

Once everything is stabilized, I can separate it into smaller PRs.


aldoborrero:
I'm closing this in favor of splitting different (but related) tasks into separate PRs.

PR #207, opened 2023-01-02
fix/default log output

### LINKED ISSUE #206: No output when syntax errors in Go file
I run the command in a dir where there's a go file which looks like this (the curly brace on a new line):

```
func main() 
{
	fmt.Println("hello world")
}
```
There's zero output, however I would expect it to forward the formatter error message to the console. Same with any other file contents not matching go, but having a .go extension.

brianmcgee:
- fix: default log level should be 'Warn' not 'Off'
- fix: log formatter error as warn instead of info

Closes #206

### DIFF
--- a/src/formatter.rs
+++ b/src/formatter.rs
@@ -1,13 +1,15 @@
 //! Utilities for the formatters themselves.
-use crate::config::FmtConfig;
-use crate::{expand_exe, expand_if_path, expand_path};
+use std::{fmt, path::Path, path::PathBuf, process::Command};
+
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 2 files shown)


brianmcgee:
Using the example in #206 the output is now:

```shell
❯ treefmt -C examples/go/
[WARN ] Error using formatter #go:
    • [STDOUT]:
    
    • [STDERR]:
    /home/brian/Development/github.com/numtide/treefmt/examples/go/main.go:6:1: unexpected semicolon or newline before {
    
[ERROR] #go's formatter failed: exit status 2
```

PR #211, opened 2023-01-14
improve the formatter load error

zimbatm:
Display what it was looking for, so it's easier to debug.

Before:

    Failed to load formatter #python due to error: cannot find binary path

After:

    Failed to load formatter #python: could not find 'black' on PATH

This should help with errors like https://github.com/numtide/treefmt-nix/issues/27


zimbatm:
bors merge

PR #212, opened 2023-01-14
change default log level to info

zimbatm:
Info is meant to inform the user. In that case, we want to show how much time each formatter is taking, because it's good for the ecosystem to be aware of how slow each formatter is.


brianmcgee (approved):
LGTM

PR #214, opened 2023-01-20
feat: clap integration

### LINKED ISSUE #194: Migrate from structop to clap

aldoborrero:
Closes #194

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -1,6 +1,6 @@
 #![allow(clippy::redundant_closure, clippy::redundant_pattern_matching)]
 
-use log::{error, LevelFilter};
+use log::error;
 use treefmt::command::{cli_from_args, run_cli};
 
 fn main() {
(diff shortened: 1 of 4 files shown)


zimbatm:
thanks!

src/main.rs
```diff
@@ -17,11 +17,7 @@ fn run() -> anyhow::Result<()> {
     env_logger::builder()
         .format_timestamp(None)
         .format_target(false)
-        .filter_level(match cli.verbosity {
-            0 => LevelFilter::Info,
-            1 => LevelFilter::Debug,
-            _ => LevelFilter::Trace,
-        })
+        .filter_level(cli.verbose.log_level_filter())
```
Mic92, line 20: With this change we don't get any errors at all. Now the loglevel is dropped to ERROR,
before it was at INFO.

PR #225, opened 2023-03-08
fix treefmt --stdin when changes are moved into tempfile

stolyaroleh:
With the following config:

```toml
[formatter.cpp]
command = "clang-format"
options = ["-i"]
includes = ["*.cpp"]
```

`treefmt --stdin foo.cpp < foo.cpp` would error:
```
[ERROR] original tempfile has been replaced at path "/home/oleh/src/treefmt/_tmpoZwnEJfoo.cpp"
```
..since `clang-format` creates its own tempfile and then moves it into our own tempfile,
replacing it and causing `tmpfile.reopen()` to fail.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -80,9 +80,9 @@ pub fn from_path(file_path: &Path) -> Result<Root> {
         .and_then(|content| from_string(&content))
 }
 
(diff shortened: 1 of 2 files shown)


zimbatm (approved):
thanks!

bors merge

PR #227, opened 2023-04-01
docs: use mkdocs-numtide

zimbatm:
Move the theme and base config to another repo so it can be shared between projects.


zimbatm:
bors merge

PR #228, opened 2023-04-06
flake: expose the list of supported systems

zimbatm:
This commit introduces a new convention that allows consumers of a flake to control systems the flake is built with.

One of the core issues of flakes is that the list of systems that a flake is evaluated with is internal to the flake. Some projects can technically build with more platforms, but to change this, the user has to fork or submit a patch to change the list. Conversely, as a consumer of a flake, you might not care of evaluating the flake with all of the systems, and that list cannot be reduced.

To work around that issue, we introduce a new convention: 
1. If the `systems` input exists, it MUST point to a nix file. 
2. That file SHOULD be named `flake.systems.nix`.
3. That file MUST contain a list of supported systems (list of strings)
4. That input MUST be used by the flake to iterate over the supported systems (see this commit as an example)

With this convention in place, consumers of the flake can now override the list of systems using the input "follows" mechanism like so:

```
inputs.systems.url = "path:./flake.systems.nix";
inputs.systems.flake = false;
inputs.treefmt.url = "github:numtide/treefmt";
inputs.treefmt.inputs.follows.systems = "systems";
```

Invented with the help of @bb010g in https://github.com/numtide/flake-utils/pull/84

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -76,7 +76,8 @@
         "flake-parts": "flake-parts",
         "mkdocs-numtide": "mkdocs-numtide",
         "nixpkgs": "nixpkgs",
-        "rust-overlay": "rust-overlay"
+        "rust-overlay": "rust-overlay",
+        "systems": "systems"
       }
     },
     "rust-overlay": {
@@ -97,6 +98,21 @@
       "original": {
         "owner": "oxalica",
         "repo": "rust-overlay",
+        "type": "github"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1680980633,
+        "narHash": "sha256-mz27VfAExPMYuoWsb1cf++DIyUWWBEbAvXD0BJ+AT/E=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "4e9a51a15ceb27e5141819142a7d2ee827943fc8",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
         "type": "github"
       }
     }
--- a/flake.nix
+++ b/flake.nix
@@ -2,6 +2,9 @@
   description = "treefmt";
   # To update all inputs:
   # $ nix flake update --recreate-lock-file
+
+  inputs.systems.url = "github:nix-systems/default";
+
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
   inputs.flake-parts.url = "github:hercules-ci/flake-parts";
   inputs.flake-parts.inputs.nixpkgs-lib.follows = "nixpkgs";
@@ -12,9 +15,9 @@
   inputs.mkdocs-numtide.url = "github:numtide/mkdocs-numtide";
   inputs.mkdocs-numtide.inputs.nixpkgs.follows = "nixpkgs";
 
-  outputs = { self, nixpkgs, flake-parts, mkdocs-numtide, ... }@inputs:
+  outputs = { self, nixpkgs, flake-parts, mkdocs-numtide, systems, ... }@inputs:
     flake-parts.lib.mkFlake { inherit self; } {
-      systems = nixpkgs.lib.systems.flakeExposed;
+      systems = import systems;
       perSystem = { system, pkgs, ... }:
         let
           packages = import ./. {


Mic92:
This one seems to be an issue:

```
$ nix flake show github:/numtide/treefmt/flake-systems
error: cannot write modified lock file of flake 'github:numtide/treefmt/flake-systems' (use '--no-write-lock-file' to ignore)
(use '--show-trace' to show detailed location information)
```

zimbatm:
Thanks, the lock file wasn't up to date for some reason. It's working for me now.

Another fun thing that this unlocks:
```
$ nix flake show github:/numtide/treefmt/flake-systems 
github:numtide/treefmt/3f505ef27a660f1f5051200103c2849b450e88ef
├───apps
│   ├───aarch64-darwin
│   ├───aarch64-linux
│   ├───x86_64-darwin
│   └───x86_64-linux
<snip>
$ echo '["x86_64-linux"]' > systems.nix
$ nix flake show github:/numtide/treefmt/flake-systems --override-input systems "path:./systems.nix"
warning: not writing modified lock file of flake 'github:numtide/treefmt/flake-systems':
• Updated input 'systems':
    'path:./flake.systems.nix?lastModified=1&narHash=sha256-FqEFZR55m1uKlWkReR%2f20uF8fph2G%2fUDeK7zzXB5rBI=' (1970-01-01)
  → 'path:./systems.nix?lastModified=1680857188&narHash=sha256-kAigsS+U8wSLkOJ79mOXliuDWLQXp%2frnUZbhQTqc3BM=' (2023-04-07)
github:numtide/treefmt/3f505ef27a660f1f5051200103c2849b450e88ef
├───apps
│   └───x86_64-linux
├───checks
│   └───x86_64-linux
<snip>
```

zimbatm:
In https://github.com/zimbatm/exp-flake-systems I tried taking the idea even further and introducing it on the flake registry level. Too bad, flake-registry only supports referencing flakes :)

srid:
> Another fun thing that this unlocks:

Can we avoid having to create `systems.nix`? What about creating a Github org, then creating per-system repos, and then doing:

```sh
nix flake show github:/numtide/treefmt/flake-systems \
  --override-input systems "github:nix-systems/aarch64-darwin"
```

🧌

srid:
> In https://github.com/zimbatm/exp-flake-systems I tried taking the idea even further and introducing it on the flake registry level. Too bad, flake-registry only supports referencing flakes :)

What if systems is an actual flake, though?

```nix
{
  outputs = _: {
    systems = [ "x86_64-linux" ];
  };
}
```

However, why not generalize all of this to be a flake-parts module?

```nix
{
  outputs = _: {
    flakeModule.systems = [ "x86_64-linux" ];
  };
}
```

cc @roberth and cf. https://github.com/hercules-ci/flake-parts/issues/25

zimbatm:
> Can we avoid having to create systems.nix?

Not for all combinations, but we might be able to provide common defaults.

The real solution would be to add built-in support to Nix itself.

bb010g:
With regards to <https://github.com/zimbatm/exp-flake-systems>, I'd really like to see a `NixOS/systems` or similar that Nixpkgs itself takes as a flake input and uses in place of their current <https://github.com/NixOS/nixpkgs/blob/db73cdeb180ba079bf91e9175fe8d72197a0156c/lib/systems/flake-systems.nix> (Nixpkgs's version of `systems`/`defaultSystems`). I do think this input needs to also provide `allSystems` (`inputs.systems + ./all.nix` if it's a directory / actual flake?) because if utilities like `flake-utils.lib.system` continue to be around and are used for validation, downstreams need to be able to override them when introducing new systems. Also, if NixOS hosted this, `allSystems` becoming out of sync with `nixpkgs.lib.platforms.all` would be a proper bug. (At the very least, I'd like to see `github:nix-community/systems` if we can't get `github:NixOS/systems`.)

zimbatm:
It's starting to take shape quite nicely. I move the experiment to https://github.com/numtide/flake-systems.

zimbatm:
Now it has its own full github org to document the pattern and host common system combos: https://github.com/nix-systems/nix-systems

zimbatm:
bors merge

PR #231, opened 2023-04-09
Don't ignore --config-file when using --stdin

### LINKED ISSUE #230: `--config-file` is ignored when using `--stdin`
**Describe the bug**

`--config-file` is ignored when using `--stdin`

**To Reproduce**

```
cd /tmp
mkdir foo bar

cd foo
treefmt --init
[INFO]: Generated treefmt template at /tmp/foo/treefmt.toml

cd ../bar
treefmt --config-file ../foo/treefmt.toml --tree-root .
[ERR]: Failed to load formatter #mylanguage due to error: cannot find binary path
[ERR]: One or more formatters are missing

cat /dev/null | treefmt --config-file ../foo/treefmt.toml --tree-root . --stdin x
[ERR]: treefmt.toml could not be found in /tmp/bar and up. Use the --init option to create one.
```

**Expected behavior**

The second invocation of `treefmt` above should give:

```
[ERROR] Ignoring formatter #mylanguage due to error: could not find 'command-to-run' on PATH
```

**System information**

```
$ uname -a
Linux io 6.2.8 #1-NixOS SMP PREEMPT_DYNAMIC Wed Mar 22 12:38:11 UTC 2023 x86_64 GNU/Linux

$ treefmt --version
treefmt 0.5.0
```

**Additional context**

lbodor:
Fix #230.

### DIFF
--- a/src/command/format_stdin.rs
+++ b/src/command/format_stdin.rs
@@ -9,6 +9,7 @@ use std::path::{Path, PathBuf};
 pub fn format_stdin_cmd(
     tree_root: &Option<PathBuf>,
     work_dir: &Path,
+    config_file: &Path,
     paths: &[PathBuf],
     selected_formatters: &Option<Vec<String>>,
 ) -> anyhow::Result<()> {
@@ -18,18 +19,6 @@ pub fn format_stdin_cmd(
             return Err(anyhow!(
             "Could not find the project directories. On Unix, check if the HOME env is missing."
         ))
-        }
-    };
-
-    // Search for the treefmt.toml from there.
-    let config_file = match config::lookup(work_dir) {
-        Some(path) => path,
-        None => {
-            return Err(anyhow!(
-                "{} could not be found in {} and up. Use the --init option to create one.",
-                config::FILENAME,
-                work_dir.display()
-            ));
         }
     };
 
--- a/src/command/format_stdin.rs
+++ b/src/command/format_stdin.rs
@@ -25,7 +25,7 @@ pub fn format_stdin_cmd(
     // Default the tree root to the folder that contains the config file
     let tree_root = tree_root.clone().unwrap_or_else(|| {
         // unwrap: since the config_file is a file, there MUST be a parent folder.
-        config_file.clone().parent().unwrap().to_path_buf()
+        config_file.parent().unwrap().to_path_buf()
     });
 
     // Check that only one path was provided
--- a/src/command/mod.rs
+++ b/src/command/mod.rs
@@ -119,32 +119,41 @@ pub fn cli_from_args() -> anyhow::Result<Cli> {
 pub fn run_cli(cli: &Cli) -> anyhow::Result<()> {
     if cli.init {
         init_cmd(&cli.work_dir)?
-    } else if cli.stdin {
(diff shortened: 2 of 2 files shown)


src/command/format_stdin.rs
```diff
@@ -21,18 +22,6 @@ pub fn format_stdin_cmd(
         }
     };
 
-    // Search for the treefmt.toml from there.
-    let config_file = match config::lookup(work_dir) {
```
lbodor, line 25: Doesn't `config::lookup()` already happen in `cli_from_args()` for both `format_cmd()` and `format_stdin_cmd()`?

src/command/format_stdin.rs
```diff
@@ -21,18 +22,6 @@ pub fn format_stdin_cmd(
...
-        Some(path) => path,
-        None => {
-            return Err(anyhow!(
-                "{} could not be found in {} and up. Use the --init option to create one.",
-                config::FILENAME,
-                work_dir.display()
-            ));
-        }
```
lbodor, line 33: This check is in `run_cli()` and now also applies to both `format_cmd()` and `format_stdin_cmd()`.

src/command/mod.rs
```diff
@@ -119,32 +119,41 @@ pub fn cli_from_args() -> anyhow::Result<Cli> {
...
     } else {
-        // Fail if configuration could not be found. This is checked
-        // here to avoid aborting before init_cmd.
-        if cli.config_file.is_none() {
-            return Err(anyhow!(
-                "{} could not be found in {} and up. Use the --init option to create one or specify --config-file if it is in a non-standard location.",
-                config::FILENAME,
-                cli.work_dir.display(),
-            ));
+        match &cli.config_file {
```
lbodor, line 123: Let me know if you prefer `is_none()` followed by `expect()`, and I'll do that instead.

zimbatm (approved):
thanks! LGTM

borg merge

src/command/format_stdin.rs
```diff
@@ -21,22 +22,10 @@ pub fn format_stdin_cmd(
...
-    };
-
     // Default the tree root to the folder that contains the config file
     let tree_root = tree_root.clone().unwrap_or_else(|| {
         // unwrap: since the config_file is a file, there MUST be a parent folder.
-        config_file.clone().parent().unwrap().to_path_buf()
+        config_file.parent().unwrap().to_path_buf()
```
lbodor, line 28: Now same as

https://github.com/numtide/treefmt/blob/d87045a6f3f6b2cbf85e1ae1eb10d548dd51829a/src/command/format.rs#L30

resolves

https://github.com/numtide/treefmt/actions/runs/4649699136/jobs/8228370543

```
error: using `clone` on a double-reference; this will copy the reference of type `&std::path::Path` instead of cloning the inner type
  --> src/command/format_stdin.rs:28:9
   |
28 |         config_file.clone().parent().unwrap().to_path_buf()
```

src/command/mod.rs
zimbatm: Both are fine by me

zimbatm:
bors merge

lbodor:
Thanks very much :ok_hand:

PR #250, opened 2023-09-23
feat: add `hidden` flag to traverse hidden files

a-kenji:
Add the `hidden` and `no-hidden` flag that allows configuring the traversal of hidden files.

Examples:

```
treefmt --hidden
```
will traverse hidden files.

```
treefmt --hidden --no-hidden
```
will not traverse hidden files, same as the default (`treefmt`).

```
treefmt --hidden --no-hidden --hidden
```
will traverse hidden files.


clippy.toml
```diff
@@ -1 +1 @@
-too-many-arguments-threshold = 9
+too-many-arguments-threshold = 10
```
a-kenji, line 1: Maybe it is time to convert the formatting options into a struct?

For example: https://github.com/numtide/treefmt/blob/654ecefd62fbf7ce5c5538fb6adcfc75e916f0c9/src/command/format.rs#L52

zimbatm: It's getting there, but it can wait

zimbatm:
Thanks, I missed this PR before cutting the release.

I don't know precisely what use-case you have in mind, but wouldn't it make more sense to add the flag in the `treefmt.toml` config file? I suspect that the setting should be project-specific.

zimbatm:
I think you're right, if it's in the config it would always format the `.git` folder as well, and that wouldn't be what we want either.

Let's just keep the command-line flag for now, since it helps with your use-case. Until we find a better way to make those distinctions as you mention with 244

a-kenji:
I have rebased on top of main in order to make use of the testing system, and added tests for the hidden flag.

The `exclude` and `gitignore` files are respected when traversing with the `--hidden` flag, but as you said the `.git` folder will be traversed.

zimbatm:
thanks!

PR #253, opened 2023-09-28
Fix flaky tests that use `std::env::set_var`

a-kenji:
Fix flaky tests that use `std::env::set_var`.

`std::env::set_var` is not safe to be used in parallel.

Add `#[ignore]` statements to these tests.
Add an extra test step to `ci.sh`, as I don't think there is a better way mark certain tests as single threaded only.

Reproducer:
```bash
for _ in {0..100}; do
  cargo t current_dir || exit 1
done
```
Will fail very fast.

Doesn't fail:
```bash
for _ in {0..100}; do
  cargo t current_dir_prefers_pwd_env_var || exit 1
  cargo t current_dir_uses_dereferenced_path_when_pwd_env_var_not_set || exit 1
done
```

Reference: https://github.com/rust-lang/rust/issues/27970

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -4,6 +4,7 @@ set -exuo pipefail
 
 # Quick sanity check
 cargo test
+cargo test -- --ignored --test-threads=1
 
 # Check that no code needs reformatting. Acts as a minimal integration test.
 cargo run -- --fail-on-change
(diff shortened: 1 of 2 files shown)


srounce:
I'm wondering how could either the tests or the implementation be improved to make it more amenable to being tested normally? The test parallelism problem was a bit of oversight on my part, though it'd be nicer to not resort to workarounds for a less than optimal initial solution and instead have a better solution.

zimbatm:
Do you have a better idea? Otherwise, I would propose to merge this PR until we find a better solution.

PR #254, opened 2023-10-07
fix: rename tree_root config environment variable to `TREEFMT_ROOT`

a-kenji:
Renames the configuration environment variable to the "new" project name.

Alternative: `TREE_ROOT`, tough I thought that probably is too generic of an identifier.

### DIFF
--- a/src/command/mod.rs
+++ b/src/command/mod.rs
@@ -61,7 +61,7 @@ pub struct Cli {
     pub work_dir: PathBuf,
 
     /// Set the path to the tree root directory. Defaults to the folder holding the treefmt.toml file.
-    #[arg(long, env = "PRJ_ROOT", default_value = ".", value_parser = parse_path)]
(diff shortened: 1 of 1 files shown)


zimbatm:
This actually came from the PRJ spec. Have a look at https://github.com/numtide/prj-spec/blob/main/PRJ_SPEC.md and tell me know what you think :)

a-kenji:
Ah, thanks! I thought it was a leftover artifact of the rename.

> Have a look at [numtide/prj-spec@main/PRJ_SPEC.md](https://github.com/numtide/prj-spec/blob/main/PRJ_SPEC.md?rgh-link-date=2023-10-07T14%3A57%3A03Z) and tell me know what you think.

I really enjoy this idea, thank you for the link. I had some similar thoughts about some sort of well-known projects based directories.

It might be good to expose somehow that `treefmt` is following this spec.

Issue #261, opened 2023-12-10
Idea on speeding up no-op scenario


KAction:
This is output of treefmt run on clean checkout of my $dayjob repository:

```
>> treefmt

traversed 3563 files
matched 1102 files to formatters
left with 0 files after cache
of whom 0 files were re-formatted
all of this in 357ms
```

And I was wondering, why it takes so slow to do nothing with less than 4k
files. Instead of doing proper profiling, I just run `strace treefmt` and
counted the syscalls. And here what I found:

1. For every formatter `treefmt` runs a bunch of `statx(2)` in attempt to
   detect if formatter changed. Such detection takes time and is inevitably
   prone to both false-positives and false-negatives.

   I suggest skipping it and leaving it to the user to bust the cache when he
   upgrades his toolchain. That would save `N * M` system calls, where `N` is
   amount of formatters and `M` is amount of entries in your $PATH.

   [link to strace snippet](https://gist.github.com/KAction/03a46c3c0fe911a3f7939d03bb73132b)

2. `treefmt` executes `statx(2)` on each file that is matched by at least one
   rule. From content of `eval-cache` I infer that file is considered
   unmodified if it has the same `mtime` and `size`.

   This is known pain point of build systems, and there is no kernel interface
   to make it any more efficient. Luckily, there is userspace daemon
   [watchman](https://github.com/facebook/watchman) that can more efficiently
   answer the question "what files were changed".

   For example, in my case it takes watchman only 20ms to tell me what files
   were changed (I intentionally changed one file):

   ```
    $ time watchman since . c:1702188097:2344484:1:60 '*.nix'
    {
        "is_fresh_instance": false,
        "files": [
            {
                "cclock": "c:1702188097:2344484:1:2",
                "new": false,
                "ctime": 1702188990,
                "dev": 66306,
                "mtime": 1702188990,
                "uid": 1000,
                "mode": 33188,
                "size": 380,
                "oclock": "c:1702188097:2344484:1:71",
                "nlink": 1,
                "ino": 27656322,
                "gid": 100,
                "exists": true,
                "name": "shell.nix"
            }
        ],
        "version": "2023.08.14.00",
        "clock": "c:1702188097:2344484:1:85",
        "debug": {
            "cookie_files": [
                "/home/kaction/devel/dayjob/backend/package/.watchman-cookie-despise-2344484-38"
            ]
        }
    }

    ________________________________________________________
    Executed in   20.58 millis    fish           external
[...]

PR #263, opened 2023-12-17
feat: Add support for using watchman(1) instead of walking filesystem

KAction:
Anecodically, for project having 4k files and 7 different formatters, this change can reduce time it takes for treefmt(1) to figure out that it has nothing to do from ~350ms down to ~55ms, out of which getting response from watchman daemon takes ~35ms, so there is room for further improvements.

If "WATCHMAN_SOCK" environment variable is not set and "watchman(1)" command are not available, "treefmt" transparently falls back on walking the file system. Setting the environment variable is highly recommended, since is avoids subprocess invocation.

Ref: https://github.com/numtide/treefmt/issues/261

Regarding further optimizations, this is what I think can be improved::

```
[DEBUG] load cache: 8.36ms (Δ 6.85ms)
...
[DEBUG] watchman query: 43.73ms (Δ 35.02ms)
[DEBUG] post-process watchman list: 50.55ms (Δ 6.82ms)
...
```


src/engine.rs
```diff
@@ -21,9 +26,43 @@ pub enum DisplayType {
...
+
+// Here "path" is the absolute path that matches "includes" patterns and was changed, but that file
+// might be need to be excluded since its parent directory matches "excludes" patterns. Hence, we
+// need to iterate parent directories.
+//
+// Filesystem walker does not need this, since it would not descend into excluded directory in the
+// first place to discover a changed file that matches "includes" pattern.
+//
+// FIXME: There should be a cleaner solution.
+fn path_excluded(gs: &GlobSet, path: &PathBuf) -> bool {
```
KAction, line 45: This one is really inefficient. For nixpkgs, I get following timings:
```
[DEBUG] watchman query: 404.46ms (Δ 120.62ms)
[DEBUG] post-process watchman list: 601.97ms (Δ 197.52ms)
```

KAction:
As for excludes, I think I can at least optimize happy path and change `excludes: Option<GlobSet>`.

KAction:
Updated the pull request. Previous version misused watchman, causing it to result way too much results.
With this version, whole nixpkgs is down to 12ms for no-op, my $dayjob repository is down to 8.5ms.

KAction:
```
error (ignored): error: SQLite database '/home/garnix/.cache/nix/eval-cache-v5/6de50559647aa37e20a3fd649274afe85c9def927c4d55696d8a84953e4ccaa3.sqlite' is busy
```

Should I be concerned?

src/command/format.rs
```diff
@@ -49,6 +52,25 @@ pub fn format_cmd(
         paths
     );
 
+    let client = match Connector::new().connect().await {
+        Err(e) => {
+            warn!(
+                "watchman is not available (err = {:?}), falling back on stat(2)",
+                e
+            );
```
zimbatm, line 60: Most users should be able to use treefmt without `watchman`, so having this as a warning isn't the best user experience for them.

I propose three modes: "auto" (default) tries watchman and silently falls back on "stat," "stat" is stat, and "watchman" errors if watchman is not available.

Then, introduce a CLI flag and/or project-level global config to allow switching between modes.

src/command/format.rs
```diff
@@ -49,6 +52,25 @@ pub fn format_cmd(
...
+            None
+        }
+        Ok(c) => {
+            // This is important. Subprocess wastes ~20ms.
+            if !std::env::var("WATCHMAN_SOCK").is_ok() {
```
zimbatm, line 65: this needs to be documented, to explain how to set it

src/formatter.rs
```diff
@@ -191,7 +201,10 @@ impl Formatter {
             options: cfg.options.clone(),
             work_dir,
             includes,
+            // TODO: Avoid clone().
```
zimbatm, line 204: the list is generally super small, I don't think the TODO needs to stay.

zimbatm (commented):
hey, I finally had the time to take a closer look and actually run it, and it's great. The only thing missing is to fix the default user experience if watchman is not installed, and a bit of documentation on how to use that feature.

KAction:
I think I addressed all suggestions it latest revision.

PR #275, opened 2024-03-14
feat: support multiple formatters with stdin

### LINKED ISSUE #274: Multiple formatters when using --stdin

I configured the formatter `dos2unix` which I essentially applied to all files in the project, then everything seemed to work fine, however in my editor I utilize `treefmt --stdin FILENAME < BUFFER` in order to have consistent formatting across the project without duplicate configuration. Recently I noticed the formatting stopped working and when I debugged this I got this:
```
$ treefmt --stdin test.nix < test.nix
[WARN ] multiple formatters matched the path. picking the first one
{a = 1; b = 2; c = 3;}
```

So it didn't format the file because `dos2unix` was the first one.

Either a way to select which formatter has priority in this case, or I guess the best case would be to run all the formatters, just like it does when you apply via `treefmt test.nix`.

- Not have multiple formatters touching the same files.
- Format file directly via argument, however my editor formatter relies on stdin/stdout formatting.

terlar:
Fixes #274


terlar:
What do you think about this change @zimbatm?

PR #278, opened 2024-04-05
fix: enable to select some formatters without `--allow-missing-formatter`

### LINKED ISSUE #210: treefmt --formatters doesn't work correctly
Executing treefmt --formatters <formatter> errors out because of other missing formatters specified in the config.

nokazn:
## Description

This PR changes checking missing formatters logic like the below.

- If some formatters are specified with `-f`/`--formatters`, check whether they are defined in config and executable.
- If not specified, check whether all formatters are executable

I think it will resolve #210.

If something goes wrong, feel free to point it out. Thank you.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -235,14 +235,16 @@ fn load_formatters(
     selected_formatters: &Option<Vec<String>>,
     stats: &mut Statistics,
 ) -> anyhow::Result<BTreeMap<FormatterName, Formatter>> {
-    let mut expected_count = 0;
     let formatter = root.formatter;
+    let expected_count = match selected_formatters {
+        Some(f) => f.len(),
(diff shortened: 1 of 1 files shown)


zimbatm:
thanks!

MangoIV:
heya! this works for me and matches how I would expect it to work, *however* somehow this invalidates the cache; e.g. 
- configure formatters `a` and `b`
- run `treefmt` 
- run `treefmt -f a` 
- run `treefmt` again, this will start reformatting with `b`
see also issue #286

MangoIV:
the issue is that my usecase is *especially* that `b` takes horrendously long, if it invalidates the cache to run the formatter without it, I can not use treefmt to run it.

PR #285, opened 2024-04-19
chore: treefmt

a-kenji:
Ran treefmt on itself.

The current main branch has failing `ci` because of it.
I have not yet looked into why it hasn't been caught prior to merging.

Edit: #282 doesn't seem to have the same checks.

### DIFF
--- a/src/engine.rs
+++ b/src/engine.rs
@@ -123,9 +123,9 @@ pub fn run_treefmt(
                 let start_time = Instant::now();
 
                 // Run the formatter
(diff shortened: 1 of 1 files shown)


infinisil:
I believe I've seen the button that certain workflows need to be approved since I was a first-time contributor. I guess this wasn't pressed before merging, and [no status checks are required for merging](https://github.com/numtide/treefmt/blob/4dd19e6e43976ddf9a3af65fe17be2f367662ed0/.github/settings.yml#L127-L128).

PR #295, opened 2024-05-17
feat/improve logging

brianmcgee:
- feat: improve formatter cache debug logging
- feat: demote some cache related logging to debug level
- fix: use resolved executable when applying a formatter
- feat: log out the formatter command being applied at debug level


format/formatter.go
```diff
@@ -80,9 +80,18 @@ func (f *Formatter) Apply(ctx context.Context, files []*walk.File, filter bool)
...
+	cmd := exec.CommandContext(ctx, f.executable, args...)
 	cmd.Dir = f.workingDir
 
+	// log out the command being executed, truncating it as required
+	if f.log.GetLevel() == log.DebugLevel {
+		cmdStr := cmd.String()
+		if len(cmdStr) >= 1024 {
+			cmdStr = cmdStr[:1024] + "..."
+		}
+		f.log.Debugf("executing: %s", cmd.String())
```
zimbatm, line 92:
```suggestion
		f.log.Debugf("executing: %s", cmdStr)
```
So it gets used :)

zimbatm: Is that the command-line arguments, or the output of the execution? If it's the former, I would argue to keep it whole instead of truncating.

zimbatm (approved):

brianmcgee:
@mergify queue

PR #296, opened 2024-05-17
feat: switch back to numtide/treefmt-nix

brianmcgee:
It now supports freeform type for settings allowing us to pass through pipeline options.


zimbatm (approved):

brianmcgee:
@mergify queue

PR #297, opened 2024-05-17
Error out when a path matches multiple formatters

brianmcgee:
When multiple formatters need to operate on the same path they must do so as part of a pipeline to avoid concurrent modification and non-deterministic issues.

This change will detect when a path has matched against more than one pipeline and error out indicating the pipelines involved.


brianmcgee:
@mergify queue

PR #301, opened 2024-05-24
feat: simplify pipeline model

brianmcgee:
For each path we determine the list of formatters that are interested in formatting it. From there, we sort
the list of formatters first by priority (lower value, higher priority) and then by name (lexicographically).

With this information we create a batch key which is based on the unique sequence of formatters. When enough paths with the same sequence is ready we apply them in order to each formatter.

By default, with no special configuration, this model guarantees that a given path will only be processed by one formatter at a time.

If a user wishes to influence the order in which formatters are applied they can use the priority field.

## TODO

- [x] profile and determine why sometimes formatting nixpkgs takes significantly longer than usual
- [x] update docs


brianmcgee:
@zowoq this is my first attempt at making it easier to configure the ordering of formatters and make it easier to handle https://github.com/nix-community/infra/pull/1266#issuecomment-2118534710

I'm keen to get your feedback.

cli/format.go
```diff
@@ -372,29 +341,46 @@ func applyFormatters(ctx context.Context) func() error {
...
+				result := priorityA - priorityB
+				if result == 0 {
+					// formatters with the same priority are sorted lexicographically to ensure a deterministic outcome
+					result = cmp.Compare(a.Name(), b.Name())
+				}
+				return result
+			})
+
+			if len(sequence) == 0 {
 				log.Debugf("no match found: %s", file.Path)
```
zimbatm, line 367: not related but this could be a good toggle to highlight unmatched files with an `--show-unmatched` flag or something like that.

cli/format.go
```diff
@@ -343,25 +325,12 @@ func applyFormatters(ctx context.Context) func() error {
...
-						return fmt.Errorf("%s failure: %w", key, err)
-					}
-					for _, path := range batch {
-						processedCh <- path
-					}
-					return nil
-				})
-			}
-		}
+	tryApply := func(task *format.Task) {
```
zimbatm, line 328: maybe 
```suggestion
	appendFile = func(file *walk.File, formatters []*Formatter)
```

cli/format.go
```diff
@@ -372,29 +341,46 @@ func applyFormatters(ctx context.Context) func() error {
...
+				priorityA := a.Priority()
+				priorityB := b.Priority()
+
+				result := priorityA - priorityB
+				if result == 0 {
+					// formatters with the same priority are sorted lexicographically to ensure a deterministic outcome
+					result = cmp.Compare(a.Name(), b.Name())
+				}
+				return result
+			})
```
zimbatm, line 364: This concern belong to the tryApply function, or a NewTask function since it's related to the stability of the key

cli/format.go
```diff
@@ -343,25 +325,12 @@ func applyFormatters(ctx context.Context) func() error {
...
-						processedCh <- path
-					}
-					return nil
-				})
-			}
-		}
+	tryApply := func(task *format.Task) {
+		// append to batch
+		key := task.Key()
+		batches[key] = append(batches[key], task)
```
zimbatm, line 331: Then here create a Batch structure with `{ Formatters: []*Formatter, Files []*File }` where you can append the file to.

zimbatm (approved):
This is great. I tried going even further in the comments but honestly it's not needed.

zowoq:
This looks good, thank you!

cli/format.go
brianmcgee: I created https://github.com/numtide/treefmt/issues/302 to track this

brianmcgee:
@zimbatm did some cleanup and updated the docs.

zimbatm (approved):

brianmcgee:
@mergify queue

PR #309, opened 2024-05-30
cli: search for the tree root by default

### LINKED ISSUE #308: treefmt2 doesn't search for the project root anymore
**Describe the bug**

When running `treefmt` in a sub-folder, it only formats that folder.

treefmt1 was walking up to find the `treefmt.toml` and then using that as the default TreeRoot variable. This was convenient because you can easily format the whole project from any place in the tree. And if you want to only format the current folder you can still run `treefmt .`

**To Reproduce**

1. ...
2. ...

**Expected behavior**

**System information**

**Additional context**

zimbatm:
Restore the treefmt 1.x behaviour where it would search for the tree root by recursively searching for the treefmt.toml file up the filesystem, starting from the current directory.

The `--tree-root-file` option will be useful to remove this bash wrapper: https://github.com/numtide/treefmt-nix/blob/2fba33a182602b9d49f0b2440513e5ee091d838b/module-options.nix#L116-L135

Fixes #308

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -13,7 +13,7 @@ type Format struct {
 	WorkingDirectory      kong.ChangeDirFlag `default:"." short:"C" help:"Run as if treefmt was started in the specified working directory instead of the current working directory."`
 	NoCache               bool               `help:"Ignore the evaluation cache entirely. Useful for CI."`
 	ClearCache            bool               `short:"c" help:"Reset the evaluation cache. Use in case the cache is not precise enough."`
-	ConfigFile            string             `type:"path" help:"Load the config file from the given path (defaults to finding treefmt.toml up)."`
+	ConfigFile            string             `type:"path" help:"Load the config file from the given path (defaults to searching upwards for treefmt.toml)."`
 	FailOnChange          bool               `help:"Exit with error if any changes were made. Useful for CI."`
 	Formatters            []string           `short:"f" help:"Specify formatters to apply. Defaults to all formatters."`
 	TreeRoot              string             `type:"path" help:"The root directory from which treefmt will start walking the filesystem (defaults to the directory containing the config file)."`
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -17,7 +17,7 @@ type Format struct {
 	FailOnChange          bool               `help:"Exit with error if any changes were made. Useful for CI."`
 	Formatters            []string           `short:"f" help:"Specify formatters to apply. Defaults to all formatters."`
 	TreeRoot              string             `type:"existingdir" xor:"tree-root" help:"The root directory from which treefmt will start walking the filesystem (defaults to the directory containing the config file)."`
-	TreeRootFile          string             `type:"path" xor:"tree-root" help:"File to search for to find the project root (if --tree-root is not passed)."`
+	TreeRootFile          string             `type:"string" xor:"tree-root" help:"File to search for to find the project root (if --tree-root is not passed)."`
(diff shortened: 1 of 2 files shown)


cli/cli.go
```diff
@@ -13,10 +13,11 @@ type Format struct {
 	WorkingDirectory      kong.ChangeDirFlag `default:"." short:"C" help:"Run as if treefmt was started in the specified working directory instead of the current working directory."`
 	NoCache               bool               `help:"Ignore the evaluation cache entirely. Useful for CI."`
 	ClearCache            bool               `short:"c" help:"Reset the evaluation cache. Use in case the cache is not precise enough."`
-	ConfigFile            string             `type:"existingfile" default:"./treefmt.toml" help:"The config file to use."`
+	ConfigFile            string             `type:"path" help:"Load the config file from the given path (defaults to finding treefmt.toml up)."`
```
brianmcgee, line 16:
```suggestion
	ConfigFile            string             `type:"path" help:"Load the config file from the given path (defaults to searching upwards for treefmt.toml)."`
```

cli/cli.go
```diff
@@ -13,10 +13,11 @@ type Format struct {
...
 	FailOnChange          bool               `help:"Exit with error if any changes were made. Useful for CI."`
 	Formatters            []string           `short:"f" help:"Specify formatters to apply. Defaults to all formatters."`
-	TreeRoot              string             `type:"existingdir" default:"." help:"The root directory from which treefmt will start walking the filesystem."`
+	TreeRoot              string             `type:"path" help:"The root directory from which treefmt will start walking the filesystem (defaults to the directory containing the config file)."`
```
brianmcgee, line 19: You can use an `xor` in kong to ensure only one of these is set. 

https://github.com/alecthomas/kong?tab=readme-ov-file#supported-tags

brianmcgee: This would allow you to restore the `existingdir` type which adds a bit extra validation

cli/format.go
```diff
@@ -69,6 +69,40 @@ func (f *Format) Run() (err error) {
...
+		}
+		Cli.ConfigFile, _, err = findUp(pwd, "treefmt.toml")
+		if err != nil {
+			return err
+		}
+	}
+
+	// search for the project root unless specified
+	if Cli.TreeRoot == "" {
+		// use the location of the treefmt.toml file by default
```
brianmcgee, line 86: I think this could be refined if the `xor` group is used for the flags

cli/format.go
```diff
@@ -384,3 +418,56 @@ func applyFormatters(ctx context.Context) func() error {
 		return nil
 	}
 }
+
+func findUp(searchDir string, fileName string) (path string, dir string, err error) {
+	for _, dir := range eachDir(searchDir) {
+		path := filepath.Join(dir, fileName)
+		if fileExists(path) {
+			return path, dir, nil
```
brianmcgee, line 426: how does this handle symlinks? Would it follow a directory symlink that might "escape" into a completely different directory tree?

brianmcgee (changes requested):

cli/format.go
zimbatm: eachDir is only decomposing the /path/ string, removing one component after the other. I don't like to follow symlinks because it can lead to infinite loops and other weirdnesses.

fileExists will return true if the file is readable. It can be a symlink to another file, that's fine.

cli/format.go
zimbatm: I don't see how

zimbatm: enlightenement gif

brianmcgee (approved):

brianmcgee:
@mergify queue

PR #314, opened 2024-06-06
cli: remove global state, init function usage

katexochen:
> magic is bad; global state is magic → no package level vars; no func init

https://peter.bourgon.org/blog/2017/06/09/theory-of-modern-go.html

### DIFF
--- a/cli/mappers.go
+++ b/cli/mappers.go
@@ -8,10 +8,8 @@ import (
 	"github.com/charmbracelet/log"
 )
 
-var Options []kong.Option
-
-func init() {
-	Options = []kong.Option{
+func NewOptions() []kong.Option {
+	return []kong.Option{
 		kong.TypeMapper(reflect.TypeOf(log.DebugLevel), logLevelDecoder()),
 	}
 }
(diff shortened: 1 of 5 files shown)


brianmcgee:
This is interesting. I come from a Java/Scala background originally, and when I first started writing Go I found myself making everything a struct like any good object-oriented programmer would do. Over time as I started looking at other projects I picked up that I shouldn't be making everything an object, hence the current style in `treefmt`. 

Having read the article, I find myself in agreement. Thanks for sharing.

I have some time today to review this, but also going forward I'll bear the article in mind.

cli/mappers.go
```diff
@@ -8,10 +8,8 @@ import (
 	"github.com/charmbracelet/log"
 )
 
-var Options []kong.Option
-
-func init() {
-	Options = []kong.Option{
+func NewOptions() []kong.Option {
```
brianmcgee, line 11: Minor nit:

```suggestion
func NewKongOptions() []kong.Option {
```

brianmcgee (approved):
Only 1 minor nit, but happy to merge without the change.

brianmcgee:
@mergifyio queue

Issue #317, opened 2024-06-11
Excluded files reported as unmatched


jaen:
**Describe the bug**

When a file is explicitly excluded from linting, be it with a `global` or formatter-specific exclude, it gets reported when running `treefmt` with a `no formatter for path`. While it's certainly useful to know if you didn't cover some file by accident, you also might not want to format certain file (say, patches) and getting reminded about them in the output is superfluous noise.

**To Reproduce**

1. set up formatting,
2. exclude a file,
3. observe `WARN format: no formatter for path: some/excluded/file` in the output.

**Expected behavior**

I would expect not being informed about files I don't plan to format. While it might make sense to warn by default for files that fall through the `includes`/`excludes` filter — say, you format/lint different files of the same type differently for whatever reason (I dunno, typo checking for different languages in your `md`-based docs), but you don't want any to be uncovered by accident — it also makes sense to never want to format certain files and not get nagged about them.

I'm not sure what's the best solution here — maybe don't want about excludes coming from `global.excludes` or maybe have an `global.ignores` config value — but anything would be welcome.

**System information**

I used nixpkgs-unstable#treefmt2 which at this point in time points to 2.0.0-rc4.

brianmcgee:
@jaen you can control the log level:

```
  -u, --on-unmatched=warn            Log paths that did not match any formatters at the specified log level, with fatal exiting the process with an error. Possible values are <debug|info|warn|error|fatal>.
```

I've set the default to `WARN` as I think it's useful. It should only output when the cache is cold. Once a file has been evaluated it shouldn't output again unless the formatters change or the config changes. 

You can demote it to `info` level with `-u info` so you'll only see it with `-v` or `debug` with `-u debug` to only see it with `-vv`.

jaen:
@brianmcgee I don't disagree it can be useful, it's just that you might _really_ want to ignore some files and then it becomes (at least lines relating to those files) noise.

I have noticed the `--on-unmatched=level` parameter, but it doesn't quite solve the issue IMO — you might still want to get informed about files you have not covered by accident, while not getting noise about files you know will never be linted. So this option is not granular enough.

I think cache also doesn't quite solve the issue, unless you are expected to actually store the cache between CI runs? Because unless you do, the cache will be cold all the time in CI and you'll get the noise for those files all the time.

[...]

PR #318, opened 2024-06-14
feat: improve unmatched logic

brianmcgee:
Separates global excludes processing from `Formatter.Wants`. This removes redundant processing of global excludes in each `Formatter.Wants` call.

If a file has been globally excluded, we do not emit an `on-unmatched` log message. This should help reduce noise as reported in #317.

### DIFF
--- a/cli/cli.go
+++ b/cli/cli.go
@@ -3,11 +3,12 @@ package cli
 import (
 	"os"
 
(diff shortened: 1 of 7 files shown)


zimbatm (approved):
:lgtm:

brianmcgee:
@mergifyio queue

PR #323, opened 2024-06-28
feat: use numtide/blueprint instead of flake.parts

brianmcgee:
Replaces [flake.parts](https://flake.parts) with [blueprint](https://github.com/numtide/blueprint)


brianmcgee:
@zimbatm the workflows are not running for the PR and I can't work out why. I made sure to preserve the `githubActions` output but for some reasons it's just not picking up any jobs to run.

zimbatm:
@brianmcgee I see all the checks green? If you want we can also switch the repo to buildbot-nix.

brianmcgee:
Ah, the checks weren't showing for me last night :shrug:

.envrc
```diff
@@ -1,6 +1,10 @@
 dotenv
+
 if ! has nix_direnv_version || ! nix_direnv_version 3.0.4; then
     source_url "https://raw.githubusercontent.com/nix-community/nix-direnv/3.0.4/direnvrc" "sha256-DzlYZ33mWF/Gs8DDeyjr8mnVmQGx7ASYqA5WlxwvBG4="
 fi
+
+watch_file flake.nix
```
zimbatm, line 7: This is already handled by the `use_flake` function, but adding it doesn't hurt and makes it more explicit.

flake.nix
```diff
@@ -29,28 +20,37 @@
...
-      ];
+    // {
+      githubActions = inputs.nix-github-actions.lib.mkGithubMatrix {
+        checks =
+          lib.getAttrs [
+            "x86_64-linux"
+            "x86_64-darwin"
+          ]
+          inputs.self.checks;
+      };
```
zimbatm, line 54: Eventually this could also be handled by blueprint. Adding a reference back to here:
https://github.com/numtide/blueprint/issues/21

nix/devshell.nix
```diff
@@ -1,59 +1,73 @@
...
-        ])
-        # include formatters for development and testing
-        (import ./formatters.nix pkgs)
-      ];
+        # docs
+        nodejs
+      ])
+      # platform dependent CGO dependencies
+      (lib.mkIf isLinux [pkgs.gcc])
+      (lib.mkIf isDarwin [pkgs.darwin.cctools])
```
zimbatm, line 36: I think this can be replaced by  `pkgs.stdenv.cc.cc`. It resolves to clang on macOS and GCC on Linux.

test/examples/nix/sources.nix
```diff
@@ -3,13 +3,13 @@ let
   #
   # The fetchers. fetch_<type> fetches specs of type <type>.
   #
-  fetch_file = pkgs: name: spec: let
-    name' = sanitizeName name + "-src";
-  in
-    if spec.builtin or true
-    then
-      builtins_fetchurl
-      {
```
zimbatm, line 12: Is it normal for this file to get reformatted?

zimbatm (approved):

nix/devshell.nix
brianmcgee: Good to know. I went ahead and removed CGO dependencies and disabled it as we don't need it anyway, this was a copy past from another project.

test/examples/nix/sources.nix
brianmcgee: I don't know how that got changed. I reverted it back, and ran `nix fmt` to be sure there wasn't an issue with the config. It didn't get formatted again. Let's keep an eye in case something is off going forward. Good catch.

brianmcgee:
@mergifyio queue

PR #325, opened 2024-07-01
fix: configure toml keys for config fields

brianmcgee:
The toml marshaller appears to be permissive with case when reading. This makes the case explicit with `toml:""` tags for everything. 


Mic92:
Works on my machine.

Mic92:
@mergify queue

Issue #327, opened 2024-07-01
Unable to create treefmt eval cache shouldn't be fatal


flokli:
**Describe the bug**

I'm running `treefmt` inside a Nix build.

Since the bump to 2.x, this started to fail, as `treefmt` fails on not being able to create some cache files.

This should be a recoverable error, potentially trying to use tmpdir instead, or just not creating the cache at all.

**To Reproduce**

1. Run `treefmt` inside a Nix derivation
2. `treefmt: error: could not resolve local path for the cache: could not create any of the following paths: /homeless-shelter/.cache/treefmt/eval-cache`
3. nonzero exit code

**Expected behavior**

Files get formatted.

**System information**

**Additional context**

flokli:
Running with `--no-cache` also doesn't help, it still fails trying to resolve the path to the cache.

flokli:
I still think the proper fix for this would be to throw a warning if the cache couldn't be configured, but then proceed without cache.

PR #328, opened 2024-07-01
fix: only try opening the cache if cache is enabled

flokli:
Otherwise `--no-cache` still fails to open the cache.

Part of #327.


brianmcgee (approved):
lgtm

brianmcgee:
@mergifyio queue

PR #329, opened 2024-07-02
README: drop unecessary meme

Mic92:
I don't think it adds much value to the project but takes up space and attention in the README.
At the end of the day, do users care what language something is written in?
One can implement a rollercoaster in Excel:
https://www.youtube.com/watch?v=IrVA1BBHFHw


Mic92:
I don't get what's wrong here. Master is also affected: https://github.com/numtide/treefmt/actions/runs/9755311877/job/26923871363

PR #330, opened 2024-07-03
feat: if we can't open the cache, fallback to no cache

### LINKED ISSUE #327: Unable to create treefmt eval cache shouldn't be fatal

brianmcgee:

Close #327

### DIFF
--- a/cli/format.go
+++ b/cli/format.go
@@ -120,8 +120,10 @@ func (f *Format) Run() (err error) {
 
 	// open the cache if configured
 	if !f.NoCache {
-		if cache.Open(f.TreeRoot, f.ClearCache, f.formatters); err != nil {
-			return err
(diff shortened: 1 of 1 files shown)


flokli (approved):

brianmcgee:
@mergifyio queue

PR #340, opened 2024-07-04
Improve summary stats

### LINKED ISSUE #337: Over-the-top precision in reporting times
**Describe the bug**

When `treefmt` reports times it seems to report somewhere between 5 and 9 decimal places beyond the seconds. This is silly. I would follow e.g. `time` and use at most 2 decimal places.

e.g.
```
INFO format | terraform: 71 files processed in 208.27347ms
```
vs
```
INFO format | terraform: 71 files processed in 208.27ms
```

Even better, use 2 decimal places when reporting seconds, and 0 decimal places when reporting ms (nobody cares about the microseconds!).

### LINKED ISSUE #339: `treefmt`'s report is confusing
Forked from discussion in https://github.com/numtide/treefmt/issues/336.

> When treefmt says it matched a file, that means 3495 files matched at least one of the includes globs for a formatter. It then applied those formatters, and afterwards only 51 files had actually changed.

That really isn't how it reads to me, for what it's worth. "formatted" reads like "ran the formatter on" not "ran the formatter on, and changed". 

Especially the "formatted X files in Y seconds" line is weird - the time is associated with all the files you ran the formatters on, not the ones that changed!

I would find something like this clearer: 
```
traversed 41932 files
emitted 41932 files for processing
formatted 34771 files (20181 changed) in 24.579394143s
```

Even better, ditch the first two lines, or demote them to verbose logging. In some circumstances I might be interested in how many files were looked at but not formatted, but not every time!

Or controversially: demote it all to verbose logging. Many formatters work like that, and indeed the information about how many files got changed and how long it took is also readily available from `git` and `time` if  you want it.

brianmcgee:
Rounds elapsed time to nearest millisecond and changes stats phrasing to make it clearer.

```console
❯ nix fmt -- -c 
traversed 97 files
emitted 97 files for processing
formatted 49 files (0 changed) in 551ms
```

Close #337
Close #339 

## TODO
- [x] update docs after https://github.com/numtide/treefmt/pull/341

### DIFF
--- a/cli/helpers_test.go
+++ b/cli/helpers_test.go
@@ -86,5 +86,5 @@ func assertStats(t *testing.T, as *require.Assertions, traversed int32, emitted 
 
 func assertFormatted(t *testing.T, as *require.Assertions, output []byte, count int) {
 	t.Helper()
-	as.Contains(string(output), fmt.Sprintf("formatted %d files", count))
+	as.Contains(string(output), fmt.Sprintf("(%d changed)", count))
 }
--- a/docs/.vitepress/config.mts
+++ b/docs/.vitepress/config.mts
@@ -26,8 +26,6 @@ versions.forEach(version => {
   sidebar[`/${version}/`] = sidebar["/"].map(({ text, link }) => ({ text, link: `/${version}${link}`}))
 })
 
-console.log("sidebar", sidebar)
-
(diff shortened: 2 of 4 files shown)


brianmcgee:
@mergifyio queue

PR #341, opened 2024-07-04
feat: simple versioned docs

brianmcgee:
It's a little brute force, but since we have a small amount of docs currently, it should help us in the near term to avoid
releasing doc updates for features that have not been released yet.

@zimbatm give it a go locally, with an emphasis on checking links behave as expected. 


zimbatm:
@brianmcgee what do you think of having a separate repo to maintain the website? Part of the release process would be to copy the docs over from the main repo.

zimbatm:
In terms of versioning of the docs, typically you only need to maintain `<major>.<minor>` versions as the tiny ones only fix bugs (if it's a bug in the docs, it then overrides the latest `<major>.<minor>`.

brianmcgee:
I agree on all points. 

I would love a branch workflow with 2.0, 2.1 and so on and during the docs build it checks out some work trees or whatnot and builds it all together.

I experimented a bit with it earlier in the week.

PR #344, opened 2024-07-05
fix: --no-cache

### LINKED ISSUE #343: You cannot combine: --no-cache --fail-on-change
**Describe the bug**

If you use `--no-cache`, the option `--fail-on-change` won't have any effect.

**To Reproduce**

1. `git init`
2. Create a file `no-cache-fail-on-change-bug.yaml`:
```yaml
key1: value1

key2: value2


key3: value3
```
3. `git add no-cache-fail-on-change-bug.yaml`
4. Configure treefmt with `yamlfmt`
5. Run `treefmt --no-cache --fail-on-change`
6. Observe no errors, but the file has changed
7. `git checkout no-cache-fail-on-change-bug.yaml`
8. Repeat steps with `treefmt --fail-on-change`
9. Observe `treefmt: error: unexpected changes detected, --fail-on-change is enabled`

**Expected behavior**
When `--fail-on-change` is used together with `--no-cache` it should work.

**System information**

Linux

```
$ treefmt --version
treefmt v2.0.2
```

brianmcgee:
Separates out the logic for detecting changes from the logic for updating the cache, fixing the case where both `--fail-on-change` and `--no-cache` are enabled.

Closes #343

### DIFF
--- a/cli/format.go
+++ b/cli/format.go
@@ -399,6 +399,8 @@ func (f *Format) detectFormatted(ctx context.Context) func() error {
 				if !(file.Info.ModTime() == currentInfo.ModTime() && file.Info.Size() == currentInfo.Size()) {
 					// record the change
 					stats.Add(stats.Formatted, 1)
+					// log the change for diagnostics
+					log.Debugf("file has been changed: %s", file.Path)
 					// update the file info
 					file.Info = currentInfo
 				}
@@ -429,7 +431,7 @@ func (f *Format) updateCache(ctx context.Context) func() error {
 		// if f.NoCache is set that means either the user explicitly disabled the cache or we failed to open on
 		if f.Stdin || f.NoCache {
 			// do nothing
-			processBatch := func() error { return nil }
+			processBatch = func() error { return nil }
 		}
 
 	LOOP:
--- a/cli/format.go
+++ b/cli/format.go
@@ -417,19 +417,19 @@ func (f *Format) updateCache(ctx context.Context) func() error {
 
 		// apply a batch
 		processBatch := func() error {
-			// if we are processing from stdin that means we are outputting to stdout, no caching involved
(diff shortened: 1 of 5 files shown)


cli/format.go
```diff
@@ -149,96 +149,21 @@ func (f *Format) Run() (err error) {
 	f.filesCh = make(chan *walk.File, BatchSize*runtime.NumCPU())
 
 	// create a channel for files that have been processed
+	f.formattedCh = make(chan *walk.File, cap(f.filesCh))
+
+	// create a channel for files that have been formatted
```
zimbatm, line 154:
```suggestion
	// create a channel for files that have been formatted
	f.formattedCh = make(chan *walk.File, cap(f.filesCh))

	// create a channel for files that have been processed
```

cli/format.go
```diff
@@ -444,6 +369,127 @@ func (f *Format) applyFormatters(ctx context.Context) func() error {
...
+				return nil
+			}
+
+			// pass the batch to the cache for updating
+			if err := cache.Update(batch); err != nil {
+				return err
+			}
+			batch = batch[:0]
+			return nil
+		}
```
zimbatm, line 433:
```suggestion
		processBatch := func() error {
			// pass the batch to the cache for updating
			if err := cache.Update(batch); err != nil {
				return err
			}
			batch = batch[:0]
			return nil
		}

		// if we are processing from stdin that means we are outputting to stdout, no caching involved
		// if f.NoCache is set that means either the user explicitly disabled the cache or we failed to open on
		if f.Stdin || f.NoCache {
			// do nothing
			processBatch := func() error { return nil }
		}
```
micro optim :-p

zimbatm (approved):
I didn't get to test it yet but it looks ok

brianmcgee:
@mergifyio queue

PR #349, opened 2024-07-09
feat/git changed paths only

### LINKED ISSUE #311: git: format only files that changed

When introducing a "dirty" project to treefmt, it can create a lot of code churn as files that previously weren't formatted are now getting changed. You often end up with a fat commit that formats the whole repo and breaks `git blame`.

When the file backend is git, only format the untrack and changed files by default.

Add a `--all-files` flag to override that default.

**Additional context**

brianmcgee:
If the modified time has not changed when compared with the git index we do not emit the file for processing.

This allows users to introduce treefmt to a repository without suffering an initial large formatting commit.

Instead, files can be formatted incrementally as they are changed.

## TODO

- [x] wait for #348 to be merged and rebase (this was branched from it)
- [ ] update docs

Closes #311


walk/walker.go
```diff
@@ -56,22 +56,22 @@ type Walker interface {
...
-func New(walkerType Type, root string, pathsCh chan string) (Walker, error) {
+func New(walkerType Type, root string, gitAllFiles bool, pathsCh chan string) (Walker, error) {
 	switch walkerType {
 	case Git:
-		return NewGit(root, pathsCh)
+		return NewGit(root, gitAllFiles, pathsCh)
 	case Auto:
-		return Detect(root, pathsCh)
+		return Detect(root, gitAllFiles, pathsCh)
 	case Filesystem:
```
zimbatm, line 65: Would it make sense to emit a warning if gitAllFiles is true in this case?

walk/git.go
```diff
@@ -89,23 +122,27 @@ func (g *gitWalker) Walk(ctx context.Context, fn WalkFunc) error {
...
 			if err != nil {
-				return err
+				return fmt.Errorf("failed to determine a relative path for %s: %w", path, err)
 			}
 
-			if _, ok := cache[relPath]; !ok {
+			if entry, ok := cache[relPath]; !ok {
 				log.Debugf("path %v not found in git index, skipping", path)
 				return nil
+			} else if !hasChanged(entry, info) {
```
zimbatm, line 138: Interesting that `entry` is available in this scope. I didn't know that the scope would expand to the else branch as well.

walk/git.go
```diff
@@ -6,40 +6,54 @@ import (
...
 	}
 
 	// cache in-memory whether a path is present in the git index
-	var cache map[string]bool
+	var cache map[string]*index.Entry
+
+	hasChanged := func(entry *index.Entry, info os.FileInfo) bool {
+		if g.allFiles {
+			// user wants all files regardless of changes
+			return true
```
zimbatm, line 51: does this means it ignores the cache?

walk/git.go
```diff
@@ -6,40 +6,54 @@ import (
...
+		}
+		// we only want to emit files that have changed when compared with the index
+		// mod time comparison is done with EPOCH (second) precision as per the POSIX spec
+		return entry.ModifiedAt.Truncate(time.Second) != info.ModTime().Truncate(time.Second)
```
zimbatm, line 55: Is the entry.ModifiedAt coming from the git commit?

brianmcgee: Yup

brianmcgee:
Closing in favour of #439

PR #354, opened 2024-07-20
fix(cli): restore PRJ_ROOT env var

### LINKED ISSUE #353: Treefmt doesn't use the `PRJ_ROOT` variable for `--tree-root`
**Describe the bug**
Treefmt doesn't use the `PRJ_ROOT` variable for `--tree-root` anymore.
This was the case prior to `v2`

**To Reproduce**
Set the `PRJ_ROOT` variable.

zimbatm:
PRJ_ROOT is coming from the https://github.com/numtide/prj-spec and was used in treefmt 1.x

Fixes #353


brianmcgee (approved):

brianmcgee:
@mergifyio queue

PR #398, opened 2024-08-28
fix: resolve commands relative to the root

zimbatm:
When configuring treefmt with relative paths to commands, for example if
you want to create your own repo-specific wrappers, treefmt would fail
if executed in a sub-directory.

This change makes sure that the relative path is resolved based on the
treefmt root instead of the current directory.


brianmcgee:
@mergifyio rebase

PR #399, opened 2024-08-30
chore: use opentofu instead of terraform

zimbatm:
Avoids the unfree shenanigans


phaer (approved):
I quickly thought about an option, but I believe it's better to do so if someone actually needs it - and removing allowUnfreePredicate is nice as well 👍

Issue #406, opened 2024-09-14
treefmt --stdin hangs if no filename is specified


jfly:
This command hangs indefinitely:

```
treefmt --stdin
```

SIGTERM doesn't stop it, I have to send a SIGKILL.

**Expected behavior**

treefmt should never hang indefinitely. In the specific case above, I'd expect to see [this error message](https://github.com/numtide/treefmt/blob/v2.0.5/cli/format.go#L213):

> only one path should be specified when using the --stdin flag

**System information**

I'm using the latest version of treefmt (at time of writing):

```shell
$ treefmt --version
treefmt v2.0.5
```

**Additional context**

- I dug into this, and we appear to be getting stuck here waiting for some goroutines (am I using that word correctly? sorry, total go noob here) to terminate: https://github.com/numtide/treefmt/blob/v2.0.5/cli/format.go#L196
- Specifically, the `applyFormatters` goroutine is stuck forever in [this for loop](https://github.com/numtide/treefmt/blob/v2.0.5/cli/format.go#L353) reading from `f.filesCh`
- The contract seems to be that the `walkFilesystem` goroutine is responsible for closing `f.filesCh`, but it exited [here](https://github.com/numtide/treefmt/blob/v2.0.5/cli/format.go#L213) *before* it [sets up a `defer` to close the channel at exit](https://github.com/numtide/treefmt/blob/v2.0.5/cli/format.go#L261-L262).
- IIUC, the fix here is as simple as moving the `defer close(f.filesCh)` to the very top of the `walkFilesystem` goroutine, so there's no chance for it to return without closing `f.filesCh`. This is a very simple change, so I'll send in a PR with the change.

PR #407, opened 2024-09-14
Avoid hanging: ensure we always close `f.filesCh`

jfly:
The contract seems to be that the `walkFilesystem` goroutine is
responsible for closing `f.filesCh`, but before this change, there were
codepaths that could result in the gorouting exiting without closing
`f.filesCh`. That shouldn't be possible anymore, so long as we keep this
statement at the top of the function =)

This fixes https://github.com/numtide/treefmt/issues/406

### DIFF
--- a/cli/format.go
+++ b/cli/format.go
@@ -201,6 +201,9 @@ func (f *Format) Run() (err error) {
 
 func (f *Format) walkFilesystem(ctx context.Context) func() error {
 	return func() error {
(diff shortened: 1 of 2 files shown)


jfly:
I just added a unittest of the old problematic behavior and confirmed that the test gets wedged without my changes.

brianmcgee:
Nice catch, thanks!

brianmcgee (approved):

PR #415, opened 2024-09-24
fix/path handling

brianmcgee:
Fixes an issue with the `git` walker, which @srounce [found](https://github.com/srounce/repro-treefmt-traversal-bug).

Tightens up how we handle path args at the same time.

- **chore: update flake.lock**
- **fix: path handling and checking git index**

### DIFF
--- a/cli/format.go
+++ b/cli/format.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strings"
 	"syscall"
 	"time"
 
@@ -34,6 +35,9 @@ var ErrFailOnChange = errors.New("unexpected changes detected, --fail-on-change 
 func (f *Format) Run() (err error) {
 	// set log level and other options
 	f.configureLogging()
+
+	// initialise stats collection
+	stats.Init()
 
 	// ci mode
 	if f.Ci {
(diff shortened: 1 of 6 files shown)


cli/format.go
```diff
@@ -35,6 +36,9 @@ func (f *Format) Run() (err error) {
 	// set log level and other options
 	f.configureLogging()
 
+	// initialise stats collection
+	stats.Init()
```
brianmcgee, line 40: Moved this earlier as it wasn't being reset correctly in tests if an error was thrown.

walk/git.go
```diff
@@ -3,16 +3,71 @@ package walk
...
-
-	"github.com/charmbracelet/log"
+	"strings"
 
 	"github.com/go-git/go-git/v5"
-	"github.com/go-git/go-git/v5/plumbing/filemode"
 )
 
+// fileTree represents a hierarchical file structure with directories and files.
+type fileTree struct {
```
brianmcgee, line 18: Simple tree structure for checking paths exist, be they real files or directories, in the git index.

walk/git.go
```diff
@@ -25,12 +80,7 @@ func (g gitWalker) Root() string {
 	return g.root
 }
 
-func (g gitWalker) relPath(path string) (string, error) {
-	// quick optimization for the majority of use cases
```
brianmcgee, line 29: Removed this as I feel it wasn't adding much and was just a source of potentially more bugs.

walk/git.go
```diff
@@ -86,51 +140,56 @@ func (g gitWalker) Walk(ctx context.Context, fn WalkFunc) error {
...
+				if err != nil {
+					return fmt.Errorf("failed to determine a relative path for %s: %w", path, err)
+				}
+
+				if !cache.hasPath(relPath) {
+					log.Debugf("path %v not found in git index, skipping", relPath)
+					return nil
+				}
 
-			return fn(&file, err)
```
brianmcgee, line 132: We were only processing the first path argument specified... :man_facepalming:

PR #431, opened 2024-10-09
feat/cobra

### LINKED ISSUE #351: Allow setting flags as config values
Original issue was created [here](https://github.com/numtide/treefmt-nix/issues/202).

In general, should we allow setting default values for cli args in the `treefmt.toml` file?

brianmcgee:
`treefmt.toml` has been extended to include an entry for most of the flags that can be passed to `treefmt`.

In addition, values can now be specified via the environment, prefixed with `TREEFMT_`.

Finally, the order of precedence for value lookup has been fixed: flag -> env -> treefmt.toml.

The docs have not been updated. I will do that in a later PR after I've finished re-shaping some other things. 

Closes #351


cmd/init/init.toml
```diff
@@ -0,0 +1,49 @@
...
+# File to search for to find the tree root (if --tree-root is not passed). (env $TREEFMT_TREE_ROOT_FILE)
+# tree-root-file = ".git/config"
+
+# Set the verbosity of logs e.g.
+# 0 = warn, 1 = info, 2 = debug (env $TREEFMT_VERBOSE)
+# verbose = 2
+
+# The method used to traverse the files within the tree root.
+# Currently supports 'auto', 'git' or 'filesystem'. (env $TREEFMT_WALK)
+# walk = "filesystem"
```
brianmcgee, line 39: @zimbatm I've disabled some values from being included in the config file. Are there any others you think shouldn't be there?

cmd/init/init.toml
```diff
@@ -0,0 +1,49 @@
...
+
+## Runs treefmt in a CI mode, enabling --no-cache, --fail-on-change and adjusting some other settings best suited to a
+## CI use case (env $TREEFMT_CI)
+# ci = true
+
+# The file into which a cpu profile will be written (env $TREEFMT_CPU_PROFILE)
+# cpu-profile = ./cpu.pprof
+
+# Exclude files or directories matching the specified globs (env $TREEFMT_EXCLUDES)
+# excludes = ["*.md", "*.gif"]
```
zimbatm, line 14: Weren't the excludes under a [global] prefix? It could be a back-compat issue.

cmd/init/init.toml
zimbatm: The only one I'm a bit under about is the "ci" one. But I also don't see it hurting.

brianmcgee: I wasn't sure about that, so I went ahead and removed it.

cmd/init/init.toml
brianmcgee: If `excludes` is not set, it falls back to `global.excludes`.

This was driven in part for simplicity, also in part because viper has trouble binding flags to subsections, or at least I couldn't see how to easily achieve it.

PR #436, opened 2024-10-11
feat/refine cache

brianmcgee:
- **feat: remove globals from stats package**
- **feat: refactor caching approach**


format/cache.go
```diff
@@ -0,0 +1,93 @@
...
+			return fmt.Errorf("failed to get formatters bucket from cache: %w", err)
+		}
+
+		// check for any newly configured or modified formatters
+		for name, formatter := range formatters {
+
+			stat, err := os.Lstat(formatter.Executable())
+			if err != nil {
+				return fmt.Errorf("failed to stat formatter executable %v: %w", formatter.Executable(), err)
+			}
```
zimbatm, line 32: How does this interact with `--allow-missing-formatter`?

format/cache.go
```diff
@@ -0,0 +1,93 @@
...
+
+		formattersBucket, err := cache.BucketFormatters(tx)
+		if err != nil {
+			return fmt.Errorf("failed to get formatters bucket from cache: %w", err)
+		}
+
+		// check for any newly configured or modified formatters
+		for name, formatter := range formatters {
+
+			stat, err := os.Lstat(formatter.Executable())
```
zimbatm, line 29: On Windows, the Lstat would fail because of the extra `.exe` file extension executables have. Nothing to fix immediately, but it's one of those places that would have to be tweaked for Windows support.

walk/cache/cache.go
```diff
@@ -0,0 +1,78 @@
...
+	var err error
+	var path string
+
+	if temp {
+		// If no cache is desired, rather than complicate the logic for cache / no cache, we just create a cache instance
+		// backed by a temporary file instead
+		if f, err := os.CreateTemp("", "treefmt-no-cache-*.db"); err != nil {
+			return nil, fmt.Errorf("failed to create a temporary db file: %w", err)
+		} else {
+			path = f.Name()
```
zimbatm, line 34: Should that file be cleaned up on exit as well?

walk/cache/cache.go
```diff
@@ -0,0 +1,78 @@
...
+		}
+	} else {
+		// Otherwise, the database will be located in `XDG_CACHE_DIR/treefmt/eval-cache/<name>.db`, where <name> is
+		// determined by hashing the treeRoot path.
+		// This associates a given treeRoot with a given instance of the cache.
+		h := sha1.New()
+		h.Write([]byte(root))
+		digest := h.Sum(nil)
```
zimbatm, line 42: Since Go 1.2 you can now do:
```suggestion
		digest := sha1.Sum([]byte(root))
```

walk/walk.go
```diff
@@ -0,0 +1,114 @@
...
+	default:
+		return nil, fmt.Errorf("unknown walk type: %v", walkType)
+	}
+
+	if err != nil {
+		return nil, err
+	}
+
+	// wrap with cached reader
+	reader, err = NewCachedReader(db, BatchSize, reader)
```
zimbatm, line 111: Wouldn't ignoring that wrap in case of `--no-cache` be simpler than creating a tempfile?

format/cache.go
brianmcgee: Before this is called `cmd/format/format.go` we process the formatters like this:

```go
	for name, formatterCfg := range cfg.FormatterConfigs {
		formatter, err := format.NewFormatter(name, cfg.TreeRoot, env, formatterCfg)

		if errors.Is(err, format.ErrCommandNotFound) && cfg.AllowMissingFormatter {
			log.Debugf("formatter command not found: %v", name)
			continue
		} else if err != nil {
			return fmt.Errorf("%w: failed to initialise formatter: %v", err, name)
		}

		// store formatter by name
		formatters[name] = formatter
	}

```

If `cfg.AllowMissingFormatter` is true then the missing formatters are omitted from the formatters we will use.

format/cache.go
brianmcgee: Good to know. I think that's something we would tackle in a specific PR targeting windows support.

walk/walk.go
brianmcgee: You're right, that logic was a hangover from how it was done. I'll update it.

cmd/format/format.go
```diff
@@ -154,12 +150,32 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
...
-			// if we can't open the cache, we log a warning and fallback to no cache
-			log.Warnf("failed to open cache: %v", err)
-			cfg.NoCache = true
+		// open the db
+		db, err = cache.Open(cfg.TreeRoot)
+		if err != nil {
+			return fmt.Errorf("failed to open cache: %w", err)
+		}
+
+		// ensure db is closed after we're finished, and removed if temporary
```
jfly, line 162: What does "temporary" refer to here? Has this comment bitrotted?

jfly (commented):
Sorry for the single nit, I started skimming this and saw a confusing comment.

I haven't read through the whole PR carefully.

cmd/format/format.go
brianmcgee: Updated it :)

zimbatm (approved):

PR #438, opened 2024-10-11
Improve path argument handling

### LINKED ISSUE #435: treefmt 2+ no longer formats gitignored files passed explicitly as arguments
**Context**

I believe treefmt 1.x was totally ignorant of VCS/git, and would happily format all files it could find (including stuff the `.git` directory!).

It's great that treefmt 2+ doesn't do this: it's default filesystem walker ("auto") detects when we're in a git repo and ignores untracked files.

**Describe the bug**

The problem is that treefmt no longer formats files that are explicitly passed as arguments.

This breaks my "generic nix fmt" formatter I wrote for null-ls/none-ls. See https://github.com/jfly/snow/blob/835b3765b2fa18768ccebb474b292c061ba47752/pkgs/neovim/vimrc#L332-L342. Much like treefmt itself, under the hood, null-ls generates tempfiles and passes them as command line parameters to the formatter. It expects the formatter to format these files, even though they're not tracked by git.

In other words, null-ls expects formatters to obey [treefmt's formatter specification](https://github.com/numtide/treefmt/blob/v2.0.5/docs/formatter-spec.md). (On a related note, I've tried to clarify this here: https://github.com/numtide/treefmt/pull/434). I'd argue that treefmt itself doesn't follow its own formatter specification, because it won't format untracked tempfiles.

(I am aware of treefmt's `--walk=filesystem` parameter, but that would require my null-ls plugin to somehow know that it's invoking treefmt, when all it really knows is that it's invoking `nix fmt`.)

**Expected behavior**

I'd like for treefmt's default behavior to be to format files passed explicitly on the command line, even if those files are not tracked by git. If folders are passed, or if `--walk=git` is used, it's fine to apply the normal filesystem walking rules.

brianmcgee:
Changes how we handle paths which are provided as arguments:

- if it's a file, we attempt to format it based on the matching options
- if it's a directory, we traverse it using the provided walk type

This is more consistent with what a user expects and makes it easier to integrate with tools such as `none-ls`.

Close #435

### DIFF
--- a/cmd/format/format.go
+++ b/cmd/format/format.go
@@ -63,49 +63,6 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 		<-time.After(time.Until(startAfter))
 	}
 
-	if cfg.Stdin {
-		// check we have only received one path arg which we use for the file extension / matching to formatters
-		if len(paths) != 1 {
-			return fmt.Errorf("exactly one path should be specified when using the --stdin flag")
-		}
-
-		// read stdin into a temporary file with the same file extension
-		pattern := fmt.Sprintf("*%s", filepath.Ext(paths[0]))
-
-		file, err := os.CreateTemp("", pattern)
-		if err != nil {
-			return fmt.Errorf("failed to create a temporary file for processing stdin: %w", err)
-		}
-
-		if _, err = io.Copy(file, os.Stdin); err != nil {
-			return fmt.Errorf("failed to copy stdin into a temporary file")
-		}
-
-		// set the tree root to match the temp directory
-		cfg.TreeRoot, err = filepath.Abs(filepath.Dir(file.Name()))
-		if err != nil {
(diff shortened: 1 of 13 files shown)


cmd/format/format.go
```diff
@@ -204,13 +161,29 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
...
 		return fmt.Errorf("invalid walk type: %w", err)
 	}
 
-	reader, err := walk.NewReader(walkType, cfg.TreeRoot, paths, db, statz)
+	if walkType == walk.Stdin {
+		// check we have only received one path arg which we use for the file extension / matching to formatters
+		if len(paths) != 1 {
+			return fmt.Errorf("exactly one path should be specified when using the --stdin flag")
+		}
+	} else {
```
zimbatm, line 175: I think this branch is also applicable if walkTyoe == walk.Stdin. Created https://github.com/numtide/treefmt/pull/440

PR #439, opened 2024-10-12
Improve git walking

brianmcgee:
Replaces the use of `go-git` with a simple `git ls-file` for traversing the git index.

## TODO
- [x] rebase after #447 is merged


brianmcgee:
@jfly wouldn't mind your eye on this too 🙏

walk/git.go
```diff
@@ -1,200 +1,143 @@
...
 ) (*GitReader, error) {
-	repo, err := git.PlainOpen(root)
-	if err != nil {
-		return nil, fmt.Errorf("failed to open git repository: %w", err)
+	// check if the root is a git repository
+	cmd := exec.Command("git", "rev-parse", "--is-inside-work-tree")
+	cmd.Dir = root
+
+	if out, err := cmd.Output(); err != nil {
+		return nil, fmt.Errorf("failed to check if git repository is inside work tree: %w", err)
```
brianmcgee, line 114: This error message doesn't make sense, need to reword it.

walk/git.go
```diff
@@ -1,200 +1,143 @@
...
+	} else if strings.Trim(string(out), "\n") != "true" {
+		return nil, fmt.Errorf("git repository is not inside work tree")
```
brianmcgee, line 116: Needs rewording

cmd/root_test.go
```diff
@@ -594,7 +585,7 @@ func TestGitWorktree(t *testing.T) {
 	// walk with filesystem instead of git
 	_, statz, err := treefmt(t, "-c", "--config-file", configPath, "--tree-root", tempDir, "--walk", "filesystem")
 	as.NoError(err)
-	assertStats(t, as, statz, 60, 60, 60, 0)
+	assertStats(t, as, statz, 80, 80, 80, 0)
```
jfly, line 588: Why did the stats change?

walk/filesystem.go
```diff
@@ -80,6 +80,11 @@ func (f *FilesystemReader) process() error {
 // Read populates the provided files array with as many files are available until the provided context is cancelled.
 // You must ensure to pass a context with a timeout otherwise this will block until files is full.
 func (f *FilesystemReader) Read(ctx context.Context, files []*File) (n int, err error) {
+	// ensure we record how many files we traversed
+	defer func() {
+		f.stats.Add(stats.Traversed, int32(n))
+	}()
+
```
jfly, line 87: Is this change related? (totally ok if it's an unrelated improvement, just make sure I understand what's going on)

walk/git_test.go
```diff
@@ -66,3 +70,85 @@ func TestGitReader(t *testing.T) {
...
+	cmd = exec.Command("git", "add", ".")
+	cmd.Dir = tempDir
+	as.NoError(cmd.Run(), "failed to add everything to the index")
+
+	// commit
+	cmd = exec.Command("git", "commit", "-m", "initial commit")
+	cmd.Dir = tempDir
+	as.NoError(cmd.Run(), "failed to commit")
+
+	// should be empty
```
jfly, line 94: What is the thing that should be empty? Is this referring to the `n` that we assert is 0 a few lines down?

I think I'm getting tripped up because the comment feels "attached" to this block of code that's constructing the `NewGitStagedReader`, and that block of code doesn't seem to be checking about the emptiness of anything.

cmd/root_test.go
```diff
@@ -537,7 +532,7 @@ func TestBustCacheOnFormatterChange(t *testing.T) {
 	assertStats(t, as, statz, 32, 0, 0, 0)
 }
 
-func TestGitWorktree(t *testing.T) {
+func TestGit(t *testing.T) {
```
jfly, line 535: If I'm reading this test correctly, it doesn't test out the situation where we have a single file with some staged changes *and* some unstaged changes. In particular, I think the scenario where the unstaged and staged changes conflict (post formatting) is most interesting.

I haven't read the implementation yet (I'm reading the tests first), so maybe this isn't actually interesting, but as a user, it *seems* interesting. Concretely, I'm thinking of something like this:

1. Create a new `test.py` file in need of some formatting:

   ```python
   print(   "woah there are a lot of spces there")
   ```

2. `git add test.py`: stage that file
3. Now edit `test.py` to fix the typo:

   ```python
   print(   "woah there are a lot of spaces there")
   ```
4. Do *not* stage that change.
5. Now invoke treefmt with `--walk=git_staged`. Does this work? It feels distressing to me because the unstaged "diff" wouldn't apply cleanly on top of the formatted staged changed. But perhaps I'm probably incorrectly assuming this stuff works on diffs. I need to go read your code now.

walk/git.go
jfly: "failed to check if we're in a git repository"?

walk/git.go
jfly: "not in a git repository"?

jfly (commented):
This is a neat idea! It'll be cool to see it in action.

I tried to give you a thorough review, but I'm a total beginner with `go`, and don't really have treefmt's architecture in my head.

That said, I think I understand enough of this PR to know that either something is missing, or I'm even more confused than I realize. Perhaps I'm just misunderstanding what this feature is meant to do.

I would have expected that if I ask treefmt to format my staged files, it would do some git acrobatics to only format the staged changes of those staged files. (You'll see this expectation reflected in at least one of my inline code comments). Now that I've read the whole diff, that doesn't seem to be the case: this really only affects which files we walk over, it doesn't do anything clever to separate the staged vs unstaged changes. Doing this right is tricky and is why writing a pre-commit hook correctly is so damn annoying.

If I understand correctly, this also doesn't seem to implement the thing asked for in https://github.com/numtide/treefmt/issues/311. But maybe the goal changed since June? I'm also a little skeptical about the thing being described in #311. I'll leave a comment over there about that.

brianmcgee:
> If I understand correctly, this also doesn't seem to implement the thing asked for in #311. But maybe the goal changed since June? I'm also a little skeptical about the thing being described in #311. I'll leave a comment over there about that.

I tagged that issue as I thought I was solving the same problem. But you're right; I seem to have drifted.

brianmcgee:
On reflection and continuing the discussion in #311, I will simplify this PR to just replacing `go-git` with `git ls-files`.

jfly:
> I will simplify this PR to just replacing `go-git` with `git ls-files`

Just curious, why are we dropping `go-git`? Coincidently, I had a conversation with @traxys a few weeks ago about an issue he was running into with some experimental git feature (`core.fsmonitor`, I think?) that (maybe?) isn't by `go-git`. I suspect that switching to `git ls-files` would have made that problem go away.

brianmcgee:
> Just curious, why are we dropping `go-git`?

Originally, this was motivated by the status in `go-git` being very slow on large repos, so when I was trying to implement the `git_staged` walk type, it was a non-starter as Nixpkgs is my reference. From there, if I was going to just shell out to `git` anyway, it felt like overkill to use `go-git`. 

Having seen how clean and simple it looks just shelling out versus using `go-git` (which I could have optimized a bit more I guess), I decided to remove an extra dependency and keep things simple. We did get burned previously by a bug in `go-git` https://github.com/nix-community/infra/pull/1266#issuecomment-2116505994. 

TLDR; less dependencies, less indirection, less code :slightly_smiling_face:

brianmcgee:
I'll rebase this once #447 is merged and then mark it ready for review.

cmd/root_test.go
```diff
@@ -733,7 +728,7 @@ func TestBustCacheOnFormatterChange(t *testing.T) {
 	})
 }
 
-func TestGitWorktree(t *testing.T) {
```
jfly, line 736: Did you mean to change the test name? Totally fine if you did, I just know this PR evolved a bit before reaching it's final form.

config/config.go
```diff
@@ -122,8 +122,8 @@ func SetFlags(fs *pflag.FlagSet) {
 	)
 	fs.String(
 		"walk", "auto",
-		"The method used to traverse the files within the tree root. Currently supports 'auto', 'git' or "+
-			"'filesystem'. (env $TREEFMT_WALK)",
+		"The method used to traverse the files within the tree root. Currently supports "+
+			"<auto|git|git_staged|git_index|filesystem>. (env $TREEFMT_WALK)",
```
jfly, line 126: Is this change vestigial?

walk/filesystem.go
```diff
@@ -80,6 +80,11 @@ func (f *FilesystemReader) process() error {
 // Read populates the provided files array with as many files are available until the provided context is cancelled.
 // You must ensure to pass a context with a timeout otherwise this will block until files is full.
 func (f *FilesystemReader) Read(ctx context.Context, files []*File) (n int, err error) {
+	// ensure we record how many files we traversed
+	defer func() {
+		f.stats.Add(stats.Traversed, int32(n))
```
jfly, line 85: Is this right? I see a loop below with cancelation logic. Is it possible that we report the wrong number if we're canceled?

walk/git.go
```diff
@@ -1,169 +1,98 @@
...
 
-			// we only want regular files, not directories or symlinks
-			if entry.Mode == filemode.Dir || entry.Mode == filemode.Symlink {
-				continue
-			}
-
-			// stat the file
-			path := filepath.Join(g.root, entry.Name)
+		// create a command which will execute from the specified sub path within root
+		cmd := exec.Command("git", "ls-files")
```
jfly, line 40: For the record, I'm reviewing this code on my phone in bed. I haven't actually checked if ls-files does what we want when run in a subdir.

walk/git.go
```diff
@@ -1,169 +1,98 @@
...
-func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error) {
-	idx := 0
+		// create a new scanner for reading the output
+		g.scanner = bufio.NewScanner(r)
+	}
 
 LOOP:
-	for idx < len(files) {
+
+	for n < len(files) {
```
jfly, line 55: I having trouble understanding this. It looks like files is an array passed into this function, and we're overwriting it with the output of git ls-files. Things I am concerned about:

- what if `files` isn't large enough?
- what if `files` is larger than the number of results we get back from `ls-files`?

Separately, are we checking the exit code of the git subproccess? Ideally we'd error out if something went wrong there.

cmd/root_test.go
```diff
@@ -799,8 +790,9 @@ func TestGitWorktree(t *testing.T) {
 	as.NoError(err)
 
 	assertStats(t, as, statz, map[stats.Type]int32{
-		stats.Traversed: 60,
-		stats.Matched:   60,
+		stats.Traversed: 80,
+		stats.Matched:   80,
```
jfly, line 794: I still don't understand why the stats have changed here.

walk/git.go
jfly: Do we have a test of using the git walker in a subdir?

cmd/root_test.go
brianmcgee: I did

config/config.go
brianmcgee: Shit, yeah.

walk/filesystem.go
brianmcgee: Wow, that logic is a bit broken and I don't know how the tests passed.

walk/git.go
brianmcgee: I'll add one if there isn't

walk/git.go
brianmcgee: Here's how we are executing `git ls-files`:

```go
if g.scanner == nil {
		// create a pipe to capture the command output
		r, w := io.Pipe()

		// create a command which will execute from the specified sub path within root
		cmd := exec.Command("git", "ls-files")
		cmd.Dir = filepath.Join(g.root, g.path)
		cmd.Stdout = w

		// execute the command in the background
		g.eg.Go(func() error {
			return w.CloseWithError(cmd.Run())
		})

		// create a new scanner for reading the output
		g.scanner = bufio.NewScanner(r)
	}
```

This does a few things:

- Creates an in memory pipe, with a reader (r) and write (w). 
- Sets the working directory for the command to `${g.root}/${g.path}`
- Sets stdout for the command to the write end of the pipe
- Runs the command in a separate go routine with `cmd.Run`, passing any error (exit code) returned into the close method of the write end of the pipe. This will emit the error on next read on the other end, nor `io.EOF` if the error was `nil`.
- We initialise a `bufio.Scanner` to read stdout line by line

The overall idea for the method is that you pass in an array with a length matching your maximum batch size. We read from the stdout of `git ls-files`, line by line, until we have filled that array, at which point we return to the user and indicate how many of them we filled `n`. This is borrowing from how other `Reader` interfaces work within go and is somewhat 'idiomatic'. 

It has the added benefit of re-using the array between subsequent calls, reducing GC.

jfly: Oh this helps! This feels very much like c-style `read` semantics.

I was very much misreading this code. TIL about go's "named return values"

cmd/root_test.go
jfly: That helps, thanks. But I'm still confused about why these numbers changed from before. Did we change what the test is doing? Or did we change what the underlying code is doing?

PR #440, opened 2024-10-12
fix: also check if path is in project for --stdin

zimbatm:
--stdin should operate under the same rules as the other walkers.

### DIFF
--- a/cmd/format/format.go
+++ b/cmd/format/format.go
@@ -172,13 +172,13 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 		if len(paths) != 1 {
 			return fmt.Errorf("exactly one path should be specified when using the --stdin flag")
 		}
-	} else {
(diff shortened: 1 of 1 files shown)


brianmcgee (approved):

brianmcgee:
I remember now why it worked the way it did before. The stdin path doesn't exist yet. It will be created by the stdin walker when `Read` is called.

jfly:
> The stdin path doesn't exist yet. It will be created by the stdin walker when Read is called.

Hmm... are you sure the stdin walker creates the file? That seems like odd behavior to me, and doesn't seem to be what's happening with the code on `main`.

https://github.com/numtide/treefmt/pull/445 removes the requirement that the paths actually exist in order to check that they're under the tree root. Assuming that PR looks good, I've re-implemented @zimbatm's work here over in https://github.com/numtide/treefmt/pull/446. I don't think it has any issues with missing files.

Issue #444, opened 2024-10-12
Error: failed to close reader when asking treefmt to format files outside of the tree root if they're specified as a relative path


jfly:
Here's what happens if I'm in some project and use a relative path to ask treefmt to format a file outside of the project:

```
proj/ $ treefmt --no-cache ../out-of-tree.py
traversed 0 files
emitted 0 files for processing
formatted 0 files (0 changed) in 2ms
Error: failed to close reader: path '/tmp/tmp.V5f56DPrFE/out-of-tree.py' is outside of the root '/tmp/tmp.V5f56DPrFE/proj'
```

This is different than nice "not found within the tree root" error message we normally get:

```
proj/ $ treefmt --no-cache /out-of-tree.py
Error: path /out-of-tree.py not found within the tree root /tmp/tmp.V5f56DPrFE/proj
```

**To Reproduce**

Here's the setup for the above commands:

```
$ cd "$(mktemp -d)"
$ touch out-of-tree.py
$ mkdir proj
$ cd proj
$ git init

$ nix shell github:numtide/treefmt nixpkgs#black
$ treefmt --version
treefmt v659aa0f
$ cat treefmt.toml   # create this file
[formatter.python]
command = "black"
includes = ["*.py"]
```

**Expected behavior**

I acknowledge this is a pretty nitty cosmetic issue. That said, I would expect treefmt to give the same error message for files outside of the tree root.

**System information**

See above. I'm using 659aa0f.

jfly:
Fix out for review here: https://github.com/numtide/treefmt/pull/445/

brianmcgee:
Fixed with #445

Issue #443, opened 2024-10-12
treefmt treats relative paths as relative to the tree root, rather than the current working directory


jfly:
treefmt refuses to format files relative to $PWD when I'm in a subdirectory of my project.

This works:

```
./ $ treefmt subdir/tracked.py
traversed 1 files
emitted 1 files for processing
formatted 1 files (0 changed) in 143ms
```

But if I'm in `subdir/`:

```
subdir/ $ treefmt ./tracked.py
Error: path ./tracked.py not found within the tree root /tmp/tmp.NFu4zPpIQ9
```

But this does work if I use the same relative path from before:

```
subdir/ $ treefmt subdir/tracked.py
traversed 1 files
emitted 0 files for processing
formatted 0 files (0 changed) in 17ms
```

**To Reproduce**

Here's the setup for the above commands:

```
$ cd "$(mktemp -d)"
$ git init
$ mkdir subdir
$ touch subdir/tracked.py
$ git add --intent-to-add subdir/tracked.py

$ nix shell github:numtide/treefmt nixpkgs#black
$ treefmt --version
treefmt v659aa0f
$ cat treefmt.toml   # create this file
[formatter.python]
command = "black"
includes = ["*.py"]
```

**Expected behavior**

I expect treefmt to interpret relative paths relative to the current working directory, *not* the tree root.

**System information**

See above. I'm using 659aa0f.

jfly:
Fix out for review here: https://github.com/numtide/treefmt/pull/445/

brianmcgee:
Fixed with #445

Issue #442, opened 2024-10-12
treefmt refuses to format files when invoked with their absolute paths


jfly:
Tracked relative paths get formatted:

```
$ treefmt tracked.py
traversed 1 files
emitted 1 files for processing
formatted 1 files (1 changed) in 107ms
```

Untracked relative path get formatted (thanks @brianmcgee! this works now because of https://github.com/numtide/treefmt/pull/438):

```
$ treefmt untracked.py
traversed 1 files
emitted 1 files for processing
formatted 1 files (0 changed) in 119ms
```

Tracked absolute paths get skipped:

```
$ treefmt $(realpath tracked.py)
Error: path /home/jeremy/tmp/2024-10-12-pyhack-1/tracked.py not found within the tree root /home/jeremy/tmp/2024-10-12-pyhack-1
```

Untracked absolute paths get skipped:
```
$ treefmt $(realpath untracked.py)
Error: path /home/jeremy/tmp/2024-10-12-pyhack-1/untracked.py not found within the tree root /home/jeremy/tmp/2024-10-12-pyhack-1
```

**To Reproduce**

Here's the setup for the above commands:

```
$ cd "$(mktemp -d)"
$ git init
$ touch tracked.py
$ touch untracked.py
$ git add --intent-to-add tracked.py

$ nix shell github:numtide/treefmt nixpkgs#black
$ treefmt --version
treefmt v659aa0f
$ cat treefmt.toml   # create this file
[formatter.python]
command = "black"
includes = ["*.py"]
```

**Expected behavior**

I'd expect treefmt to format these files.

**System information**

See above. I'm using 659aa0f.

jfly:
Fix out for review here: https://github.com/numtide/treefmt/pull/445/

brianmcgee:
Fixed with #445

PR #445, opened 2024-10-12
fix: correct handling of absolute and relative paths

jfly:
This fixes a few issues with both relative and absolute paths:

- Previously, treefmt refused to format any files passed as absolute paths, regardless of if they were in tree or not: https://github.com/numtide/treefmt/issues/442
- Previously, treefmt would treat relative paths as relative to the project root, which is not ideal if you're in a subdirectory and you just want to format the file you're right next to: https://github.com/numtide/treefmt/issues/443
  - There was even a test asserting this behavior. I changed it to reflect the updated behavior.
- treefmt's handling of paths outside of the project root was a bit inconsistent: https://github.com/numtide/treefmt/issues/444

### DIFF
--- a/cmd/format/format.go
+++ b/cmd/format/format.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"runtime/pprof"
+	"strings"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


cmd/format/format.go
```diff
@@ -173,11 +174,27 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
...
+			if err != nil {
+				return fmt.Errorf("error computing absolute path of %s: %w", path, err)
+			}
+
+			relativePath, err := filepath.Rel(cfg.TreeRoot, absolutePath)
+			if err != nil {
+				return fmt.Errorf("error computing relative path from %s to %s: %s", cfg.TreeRoot, absolutePath, err)
+			}
+
+			if strings.HasPrefix(relativePath, "..") {
```
jfly, line 190: I honestly don't know if there are issues with this check. I know it would fail to detect paths that start with `/`, but I think that can't happen right after we've called `filepath.Rel`.

This still seems too simple to be right. But we're also just trying to protect people from themselves here, I don't think there's a security issue. It's not like we're not Apache and need to be hyper careful not to serve up files outside of the webroot directory.

brianmcgee: I agree

brianmcgee (approved):
This is a good catch. I've been re-shaping things a fair bit this week and I think at some point this was on my list of things to bear in mind, but evidently it didn't make the final cut.

PR #446, opened 2024-10-12
fix: normalize and enforce stdin paths as well

jfly:
See https://github.com/numtide/treefmt/pull/440 for the last attempt to
do this because (at the time) this "is the file in the tree root?" check
relied upon checking for concrete files in the filesystem. As of
https://github.com/numtide/treefmt/commit/ff3de21a2b35a3218e845609ccd5a6a7215e9fdc,
that's longer the case, so we can make this code a bit more consistent.


brianmcgee (approved):

brianmcgee:
@jfly rebase and then mark this as ready and we can get it merged

PR #447, opened 2024-10-13
feat: match before checking cache

### LINKED ISSUE #433: treefmt cache results in warnings getting suppressed
**Describe the bug**

treefmt helpfully logs warnings when it doesn't know how to format a file:

```
$ treefmt test.py
WARN format: no formatter for path: test.py
traversed 1 files
emitted 1 files for processing
formatted 0 files (0 changed) in 21ms
```

However, if you try to format that exact same file again, you won't see the warning:

```
$ treefmt test.py
traversed 1 files
emitted 0 files for processing
formatted 0 files (0 changed) in 13ms
```

Note how the warning reappears if you invoke treefmt with `--no-cache`:

```
$ treefmt --no-cache test.py
WARN format: no formatter for path: test.py
traversed 1 files
emitted 1 files for processing
formatted 0 files (0 changed) in 3ms
```

**To Reproduce**

Run treefmt twice on the same file.

**Expected behavior**

I'd like to see the same set of warnings every time, regardless of the state of my treefmt cache.

**System information**

I'm using the the latest code on `main` (714cf33489c936d118d60dcbc066b79a53b8b271 at time of writing).

brianmcgee:
This changes the flow of processing to make unmatched behaviour more consistent.

Before, we had been:

- traversing the filesystem
- comparing with the cache and only emitting files which had changed
- applying the matching rules to determine which formatters should be applied to a given file
- applying the formatters

Now, we do the following:

- traverse the filesystem
- apply the matching rules to determine which formatters should be applied to a given file
- compare with the cache and only emit files which have changed for formatting
- apply the formatters

It does mean we are applying the matching rules against files which we may not have to format, but in testing against Nixpkgs the performance impact appears negligible.

This makes sense since most of the processing time will be spent in the formatters, not applying some globs to file paths.

Close #433


brianmcgee:
I just realized I can further simplify the stats since `matched` and `formatted` will always be the same.

jfly (approved):
This looks great, thanks for making the change =)

> I just realized I can further simplify the stats since matched and formatted will always be the same.

Sorry, I think I'm not understanding the vocabulary. Here's what I thought:

- matched: the number of files in the repo that treefmt "has opinions" about
- formatted: of the files we matched, the number of files that treefmt invoked a formatter on. caching means we may choose not to format a matched file (because we know it hasn't changed).
- changed: of the files we formatted, how many of them actually materially changed

brianmcgee:
> * matched: the number of files in the repo that treefmt "has opinions" about
> * formatted: of the files we matched, the number of files that treefmt invoked a formatter on. caching means we may choose not to format a matched file (because we know it hasn't changed).
> * changed: of the files we formatted, how many of them actually materially changed

This is a good summary.

jfly:
Sorry, I didn't get to my point, haha:

> since matched and formatted will always be the same.

Is this true? When we have cache hits, won't we have files that we match that we end up not formatting?

cmd/format/format.go
```diff
@@ -205,33 +205,34 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
...
+		// ensure context is cancelled to release resources
+		cancel()
 
-			// check if this file is new or has changed when compared to the cache entry
-			if file.Cache == nil || file.Cache.HasChanged(file.Info) {
-				filesCh <- file
-				statz.Add(stats.Emitted, 1)
-			}
+		// pass each file into the file channel for processing
+		for idx := 0; idx < n; idx++ {
```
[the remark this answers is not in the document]
brianmcgee: The `applyFormatters` goroutine has state:

- an `errgroup.Group` limited to `runtime.NumCPU()` for executing the formatting tasks
- a map of `batch key -> []*format.Task` which is used to group formatting tasks by the unique ordered set of formatters to be applied. 

We feed it files via the channel, it assigns them to the correct batch based on the batch key, and then triggers a format if the batch reaches the `BatchSize`, flushing any partial batches when the file channel is closed. 

We want the above logic to be happening concurrently with traversing the file system, invoking formatters as soon as we can. The batch size is set to `1024` at the moment, so most of the time the formatters are getting invoked after we close the channel. But for larger repos it will have a pipeline effect. 

There's also #335 (which needs rebased) to allow for finer grained control of the batch size.

brianmcgee:
> Sorry, I didn't get to my point, haha:
> 
> > since matched and formatted will always be the same.
> 
> Is this true? When we have cache hits, won't we have files that we match that we end up not formatting?

You're not wrong. Now that I look at this again, I have no idea why I removed the distinction. I think I need to take a break from treefmt for a few days after this. 

Putting it back in.

cmd/format/format.go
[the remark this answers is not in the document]
jfly: Ah! This really helps me understand the architecture, thank you.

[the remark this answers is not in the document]
jfly: Also, for a little more context, when I wrote this comment this morning, I (incorrectly) thought that checking if a file is cached is a moderately expensive operation (I was thinking we had to read the whole file to compute a hash). I have since read the relevant cache query code and I see that it's just based on time + filesize.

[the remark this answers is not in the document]
brianmcgee: Yeah, we may eventually make the comparison strategy configurable and allow users to decide how expensive they want it to be. But since its inception, this is the heuristic `treefmt` has used. 

There is an issue that discusses the pros and cons of different approaches; I'll try to find it.

cmd/format/format.go
```diff
@@ -205,33 +205,34 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
...
 	for {
+		// read the next batch
 		ctx, cancel := context.WithTimeout(ctx, 1*time.Second)
-
 		n, err := reader.Read(ctx, files)
 
-		for idx := 0; idx < n; idx++ {
-			file := files[idx]
+		// ensure context is cancelled to release resources
+		cancel()
```
jfly, line 217: For th erecord, I do not know enough about go to understand what this is about. I trust you, though =)

cmd/format/format.go
```diff
@@ -344,21 +352,30 @@ func applyFormatters(
...
+
 				log.Logf(unmatchedLevel, "no formatter for path: %s", file.RelPath)
-				// mark it as processed and continue to the next
-				formattedCh <- &format.Task{
-					File: file,
+
+				// release the file
+				if err := file.Release(); err != nil {
+					return fmt.Errorf("failed to release file: %w", err)
 				}
```
jfly, line 367: This looks a lot like the file release code up above. Is there some way we can restructure this `for` loop so we can ensure the file is always released, regardless of the codepath? Could go's `defer` statement help here?

(for the record, I don't know what "releasing" a file means, nor why it's something we didn't do before but have to do now)

jfly (approved):
LGTM!

cmd/format/format.go
brianmcgee: You have to ensure `cancel()` is called on a context. Usually, it's done within a `defer`, but you can't do that within a `for` loop. `defer` is bound to a function.

cmd/format/format.go
brianmcgee: `defer` is bound to a function and can't be used within a `for` loop. I'll try restructuring the logic to make it clearer and require only one release block. 

FYI: I'm playing with the concept of releasing in the layered `walk.Reader`. Any reader implementation can append a hook that gets run when `release` is called. Currently, the cached reader is using that to update the cache, and the stdin reader uses it to delete the temporary file it created.

brianmcgee: I restructured it like this and actually caught a missing release in the process:

https://github.com/numtide/treefmt/blob/52ac241979fd34dc4f3b0a489977bb928d184ade/cmd/format/format.go#L334-L383

jfly: Cool! The restructuring looks good to me, thanks. Glad it caught a bug!

My poor python brain is distressed to see cleanup code written so linearly. I'm very hard coded to assume that any line of code can throw an exception, and I need to use a `try...finally` or a `with` to ensure things actually happen. I guess that's not necessary in go because there are no implicit goto exceptions, just explicit required error handling?

For the record, it's still hard for me to convince myself we actually always release these files, because I don't see it happening here for the files we matched. Perhaps there's a slightly different structure to all of this that would make even that clear?

(don't take this as actionable feedback, I'm just thinking out loud).

brianmcgee: After I merge the in-flight PRs, I'll update the `golangci-lint` config to match https://github.com/nix-community/go-nix and do a general cleanup. Then, I want to focus on making the workflow easier to follow, as, like you, I'm not 100% comfortable with it right now.

PR #450, opened 2024-10-14
fix: do not cache formatting errors

### LINKED ISSUE #449: Do not cache files that were part of a batch that had errors when formatting
Recent changes on main removed this logic:

```go
// Append to batch and process if we have enough.
// We do not cache any files that were part of a pipeline in which one or more formatters failed.
// This is to ensure those files are re-processed in later invocations after the user has potentially
// resolved the issue, e.g. fixed a config problem.
if len(task.Errors) == 0 {
	batch = append(batch, task)
	if len(batch) == BatchSize {
		if err := processBatch(); err != nil {
			return err
		}
	}
}
```

brianmcgee:
This was a regression introduced when the new `walk.Reader` approach was applied.

When a formatter errors out during processing, we should not update those files in the cache to ensure they are retried during later invocations.

I've extended the `walk.ReleaseFunc` to accept an optional `error` which can be used to indicate the file participated in a batch that had errors when being formatted.

This allows the hook implementation to decide what it should do.

Closes #449


cmd/root_test.go
```diff
@@ -473,6 +473,74 @@ func TestCache(t *testing.T) {
...
+			"fail": {
+				Command:  "touch",
+				Options:  []string{"--bad-arg"},
+				Includes: []string{"*.py"},
+			},
+		},
+	}
+	test.WriteConfig(t, configPath, cfg)
+
+	// formatters have changed, so the cache is busted and we start fresh
```
jfly, line 494: I'm confused by this comment. If we're starting fresh, why aren't we erroring out trying to format `*.py` files?

cmd/root_test.go
```diff
@@ -473,6 +473,74 @@ func TestCache(t *testing.T) {
...
+	as.NoError(err)
+
+	assertStats(t, as, statz, map[stats.Type]int32{
+		stats.Traversed: 32,
+		stats.Matched:   8,
+		stats.Formatted: 6,
+		stats.Changed:   0,
+	})
+
+	// running again with a hot, the two python files will fail and the haskell files will be skipped format as they
```
jfly, line 505: Is this comment accurate? It looks like this command does not error out.

cmd/root_test.go
```diff
@@ -473,6 +473,74 @@ func TestCache(t *testing.T) {
...
+	as.NoError(err)
+
+	assertStats(t, as, statz, map[stats.Type]int32{
+		stats.Traversed: 32,
+		stats.Matched:   8,
+		stats.Formatted: 2,
+		stats.Changed:   2,
+	})
+
+	// with a hto cache nothing and good config nothing should be formatted or changed this time
```
jfly, line 534: spelling nit:

```suggestion
	// with a hot cache nothing and good config nothing should be formatted or changed this time
```

walk/cached.go
```diff
@@ -24,8 +24,9 @@ type CachedReader struct {
 	delegate Reader
 
 	eg *errgroup.Group
-	// releaseCh contains files which have been released after processing and can be updated in the cache.
-	releaseCh chan *File
+
+	// updateCh contains files which have been released after processing and can should be updated in the cache.
```
jfly, line 28: "can should"? ;)

walk/cached.go
```diff
@@ -95,8 +96,13 @@ func (c *CachedReader) Read(ctx context.Context, files []*File) (n int, err erro
 			}
 
 			// set a release function which inserts this file into the release channel for updating
```
jfly, line 98: Has this comment bit-rotted now?

jfly (approved):
Code LGTM. I'm not totally understanding the tests, but I'm giving you a shipit because I trust that things are correct!

cmd/root_test.go
brianmcgee: I'll clear up the comments in that test and make it clearer what it's expecting to happen

cmd/root_test.go
```diff
@@ -473,6 +473,58 @@ func TestCache(t *testing.T) {
...
+	as.NoError(err)
+
+	assertStats(t, as, statz, map[stats.Type]int32{
+		stats.Traversed: 32,
+		stats.Matched:   6,
+		stats.Formatted: 0,
+		stats.Changed:   0,
+	})
+
+	// let's fix the python config so it no longer fails
```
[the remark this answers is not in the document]
brianmcgee: yeah, updated

cmd/root_test.go
```diff
@@ -473,6 +473,58 @@ func TestCache(t *testing.T) {
...
+				Options:  []string{"--bad-arg"},
+				Includes: []string{"*.hs"},
+			},
+		},
+	}
+	test.WriteConfig(t, configPath, cfg)
+
+	// running should match but not format anything
+	_, statz, err = treefmt(t, "--config-file", configPath, "--tree-root", tempDir)
+	as.NoError(err)
```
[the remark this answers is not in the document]
brianmcgee: https://github.com/numtide/treefmt/pull/366 introduced the idea that treefmt would apply the formatters on a best-effort basis. But what appears to have been missed is that the process should exit with an error if any errors are thrown. 

I've created https://github.com/numtide/treefmt/issues/451 to track this.

PR #452, opened 2024-10-15
feat: more stringent golangci-lint rules

brianmcgee:
Applies more stringent `golangci-lint` rules to the repository, borrowed from https://github.com/nix-community/go-nix.

Also fixes a bug where we were not returning errors when calling `walk.Reader.Read()` in `cmd/format`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,0 +1,39 @@
+# taken from https://github.com/nix-community/go-nix/blob/main/.golangci.yml
+linters:
+    enable:
+        - errname
+        - exhaustive
+        - gci
+        - gochecknoglobals
+        - gochecknoinits
+        - goconst
+        - godot
+        - gofumpt
+        - goheader
+        - goimports
+        - gosec
+        - importas
+        - ireturn
+        - lll
+        - makezero
+        - misspell
+        - nakedret
+        - nestif
+        - nilerr
+        - nilnil
+        - nlreturn
+        - noctx
+        - nolintlint
+        - prealloc
+        - predeclared
+        - revive
+        - rowserrcheck
+        - stylecheck
+        - tagliatelle
+        - tenv
+        - testpackage
+        - unconvert
+        - unparam
+        - wastedassign
+        - whitespace
+        - wsl
--- a/build/build.go
+++ b/build/build.go
@@ -1,6 +1,6 @@
 package build
 
 var (
-	Name    = "treefmt"
-	Version = "v0.0.1+dev"
+	Name    = "treefmt"    //nolint:gochecknoglobals
+	Version = "v0.0.1+dev" //nolint:gochecknoglobals
 )
--- a/cmd/format/format.go
+++ b/cmd/format/format.go
@@ -14,19 +14,17 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/numtide/treefmt/walk/cache"
-	bolt "go.etcd.io/bbolt"
-
 	"github.com/charmbracelet/log"
 	"github.com/gobwas/glob"
 	"github.com/numtide/treefmt/config"
 	"github.com/numtide/treefmt/format"
 	"github.com/numtide/treefmt/stats"
 	"github.com/numtide/treefmt/walk"
+	"github.com/numtide/treefmt/walk/cache"
 	"github.com/spf13/cobra"
 	"github.com/spf13/viper"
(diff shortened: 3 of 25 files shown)


.golangci.yml
```diff
@@ -0,0 +1,38 @@
+linters:
```
jfly, line 1: Is this `treefmt`'s list? If this is someone else's list we're aiming to keep in sync with, let's add a link?

build/build.go
```diff
@@ -1,6 +1,8 @@
 package build
 
 var (
-	Name    = "treefmt"
+	//nolint:gochecknoglobals
+	Name = "treefmt"
+	//nolint:gochecknoglobals
 	Version = "v0.0.1+dev"
```
jfly, line 7: Does this affect only the next line? https://golangci-lint.run/usage/false-positives/#nolint-directive says this is "to exclude issues for the block of code". I'm not sure what "block of code" means here. Is it the entire `var(...)` directive?

I personally would write these as inline hints, which the docs clearly say "excludes issues only for this line."

```suggestion
	Name = "treefmt" //nolint:gochecknoglobals
	Version = "v0.0.1+dev" //nolint:gochecknoglobals
```

cmd/format/format.go
```diff
@@ -243,7 +245,7 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 	return reader.Close()
 }
 
-// applyFormatters
+// applyFormatters.
```
jfly, line 248: Lol this dot looks funny. Is the linter enforcing that we write actual sentences? Perhaps better to just remove this comment?

cmd/init/init.go
```diff
@@ -12,9 +12,11 @@ import (
 var initBytes []byte
 
 func Run() error {
-	if err := os.WriteFile("treefmt.toml", initBytes, 0o644); err != nil {
+	if err := os.WriteFile("treefmt.toml", initBytes, 0o600); err != nil {
```
jfly, line 15: Is this really a linting related change? Feels like a change in behavior to me.

format/cache.go
```diff
@@ -25,19 +26,18 @@ func CompareFormatters(db *bolt.DB, formatters map[string]*Formatter) error {
...
 			entry, err := formattersBucket.Get(name)
-			if err != nil {
+			if !(err == nil || errors.Is(err, cache.ErrKeyNotFound)) {
 				return fmt.Errorf("failed to retrieve cache entry for formatter %v: %w", name, err)
 			}
 
-			isNew := entry == nil
-			hasChanged := entry != nil && !(entry.Size == stat.Size() && entry.Modified == stat.ModTime())
+			isNew := errors.Is(err, cache.ErrKeyNotFound)
+			hasChanged := !(isNew || (entry.Size == stat.Size() && entry.Modified == stat.ModTime()))
```
jfly, line 40: uber nit/style thing/feel free to disagree/ignore: I think we could rework this a little to be cleaner? Would this work/appease the linter?

```go
entry, err := formattersBucket.Get(name)

isNew := errors.Is(err, cache.ErrKeyNotFound)
if isNew {
    err = nil
}

if err != nil {
    ...
}

...
```

walk/cache/bucket.go
```diff
@@ -23,12 +25,14 @@ func (b *Bucket[V]) Size() int {
 func (b *Bucket[V]) Get(key string) (*V, error) {
 	bytes := b.bucket.Get([]byte(key))
 	if bytes == nil {
-		return nil, nil
+		return nil, ErrKeyNotFound
```
jfly, line 28: For my own go knowledge: why did the linter want this to return an error here? It feels pretty reasonable to me for a `get` function to decide if it wants to be "strict" or not.

It does feel to me like this change has complicated the call sites.

jfly (approved):
I am definitely not knowledgeable enough about go to give good feedback here.

I didn't see anything scary, though!

walk/cache/bucket.go
brianmcgee: It came from this linter https://github.com/Antonboom/nilnil

> return nil, nil is not idiomatic for Go. The developers are used to the fact that if there is no error, then the return value is valid and can be used without additional checks:

cmd/format/format.go
brianmcgee: Yeah I'll just remove it

PR #453, opened 2024-10-15
fix: exit with error if any formatters fail

brianmcgee:
We try to apply formatters to all files on a best-effort basis, continuing if formatting any particular batch of files fails.

This fix ensures that if any formatting errors occur, the process will exit with an error.

Closes #450

## TODO 
- [x] Merge #452 first


jfly (approved):
LGTM!

Issue #456, opened 2024-10-17
Should we expire paths?


brianmcgee:
Currently, if the formatter config never changes and the cache is not manually cleared, over time, we will start to accrue paths in the cache, which may have been removed from the underlying filesystem. 

Should we implement some form of path expiry?

jfly:
IIUC, I believe this could get worse if we implement https://github.com/numtide/treefmt/issues/455.

Not saying that it's bad right now, or that "worse" is bad either. For all I know, this file is, and always will be pretty small. I'll be keep an eye in it on my machine.

PR #457, opened 2024-10-17
feat: improve change detection

### LINKED ISSUE #455: Simplify formatter change detection
If formatters have changed, we bust the cache. Currently, that is done by checking if any formatters are added or removed and by checking their executable size and mod time. 

This could be simpler (e.g. a simple hash of everything) and encompass more information, such as formatter config.

brianmcgee:
Previously, we were storing the last `modtime` and `size` for a given path in `boltdb`.
This was used to determine if a file had changed and whether we should format it.

In addition, if a formatter's executable changed (`modtime` or `size`), we would delete
all path entries in the database before processing, thereby forcing each file to be 
formatted.

This commit introduces a new approach to change detection, one which takes into account
changes to the underlying file, the formatter's executable, _and_ the formatter's
configuration.

Now, when deciding if we should format a file, we do the following:

- Hash each matching formatter, in sequence, using its `name`, `options`, `priority`
  as well as the `modtime` and `size` of its executable.
  This is pre-computed on a per-pipeline basis.
- We then add the file's `modtime` and `size` to generate a format signature.
  You can think of this signature as a unique representation of what we are about to do
  with the file.
- The format signature is then compared with a cache entry (if available).
- If the signatures match, we have already applied this sequence of formatters, with these
  particular options etc. to this file when it had this `modtime` and `size`, so there is
  no more processing to be done.
- If the signatures do no match, we should format the file, recording the new format
  signature in the cache when we are finished.

This approach is simpler in terms of storage, and has the added benefit of finer grained
change detection versus the brute force cache busting we were doing before.

In terms of performance impact, with the pre-computing of hashes per-pipeline and the 
simpler storage schema, there appears to have been no significant impact. 
Manual testing with [nixpkgs](https://github.com/nixos/nixpkgs) shows comparable run times
for both hot and cold caches.

> [!NOTE]
> Since this changes the database schema, rather than implementing some form of migration 
> logic to remove the old buckets and so on, I decided to upgrade the hash algorithm we 
> use when determining the filename for the db file. 
> 
> Previously, we were using `sha1`, matching the behaviour from `1.0`. 
> Now we use `sha256`, which results in a slightly longer db name, but has the benefit of 
> ensuring a new db instance will be created on first invocation, as well as making 
> [gosec](https://golangci-lint.run/usage/linters/#gosec) happy.

Closes #455


format/composite.go
```diff
@@ -345,6 +263,84 @@ func (c *CompositeFormatter) Close(ctx context.Context) error {
...
+		bucket, err := tx.CreateBucketIfNotExists(bucketKey)
+		if err != nil {
+			return fmt.Errorf("failed to create formatters bucket: %w", err)
+		}
+
+		// load the previous hash which might be nil
+		prevHash := bucket.Get(hashKey)
+
+		c.log.Debug(
+			"comparising formatter hash with db",
```
jfly, line 319: Nit: "comparing"?

format/composite.go
```diff
@@ -345,6 +263,84 @@ func (c *CompositeFormatter) Close(ctx context.Context) error {
...
+		}
+	}
+
+	// finalize
+	return hex.EncodeToString(h.Sum(nil)), nil
+}
+
+// BustCache compares the current Hash() of this formatter with the last entry stored in the db.
+// If it has changed, we remove all the path entries, forcing a fresh cache state.
+func (c *CompositeFormatter) BustCache(db *bolt.DB) error {
```
jfly, line 298: I'm surprised to still see a function like this. Can't we just not do this work, and the right thing will happen?

Is this just a mechanism for keeping the cache small? Or is this a correctness thing?

If this is just a mechanism for keeping the cache small, I'd rather drop it in favor of implementing #456.

format/composite.go
```diff
@@ -345,6 +263,84 @@ func (c *CompositeFormatter) Close(ctx context.Context) error {
...
+// Hash takes anything that might affect the paths to be traversed, or how they are traversed, and adds it to a sha256
+// hash.
+// This can be used to determine if there has been a material change in config or setup that requires the cache
+// to be invalidated.
+func (c *CompositeFormatter) Hash() (string, error) {
+	h := sha256.New()
+
+	// start with the global excludes
+	h.Write([]byte(strings.Join(c.cfg.Excludes, " ")))
+	h.Write([]byte(strings.Join(c.cfg.Global.Excludes, " ")))
```
jfly, line 275: I don't understand why the excludes matter for caching. I'd expect those to only affect filesystem walking, which IIUC, is not something we cache.

For example:

1. My project has 2 files: `a.py` and `b.py`. `b.py` is currently excluded.
2. I format my whole project. I get a cache entry for `a.py`.
3. I then change my excludes so `b.py` is now included.
4. I format my whole project. I see a cache entry for `a.py`, so I have nothing to do there. I don't see a cache entry for `b.py`, so I format it and add an entry.

format/formatter.go
```diff
@@ -52,6 +54,28 @@ func (f *Formatter) Executable() string {
 	return f.executable
 }
 
+// Hash adds this formatter's config and executable info to the config hash being created.
+func (f *Formatter) Hash(h hash.Hash) error {
+	h.Write([]byte(f.name))
+	h.Write([]byte(strings.Join(f.config.Options, " ")))
+	h.Write([]byte(strings.Join(f.config.Includes, " ")))
+	h.Write([]byte(strings.Join(f.config.Excludes, " ")))
```
jfly, line 62: As mentioned above, I'm surprised to see excludes/includes here. I'd only expect us to hash things that would affect the formatting operation on a single file.

format/formatter_test.go
```diff
@@ -47,3 +52,227 @@ func TestInvalidFormatterName(t *testing.T) {
 		as.ErrorIs(err, format.ErrInvalidName)
 	}
 }
+
+func TestFormatterHash(t *testing.T) {
```
jfly, line 56: I feel like this test would get a lot shorter if we introduced a helper method. Something like:

```go
func assertHashChangeAndStable(f *formatter, oldHash int) (newHash int) {
	newHash, err := f.Hash()
	as.NoError(err)

	as.NotEqual(oldHash, newHash, "hash should have changed")

	// hash again without making changes
	newHash2, err := f.Hash()
	as.NoError(err)
	as.Equal(newHash, newHash2, "hash should not have changed")

	return newHash
}
```

test/test.go
```diff
@@ -60,13 +62,20 @@ func TempFile(t *testing.T, dir string, pattern string, contents *string) *os.Fi
...
 	t.Helper()
 
-	src, err := os.Readlink(path)
+	var utimes [2]unix.Timeval
+	utimes[0] = unix.NsecToTimeval(atime.UnixNano())
+	utimes[1] = unix.NsecToTimeval(mtime.UnixNano())
 
-	require.NoError(t, err, "failed to read symlink")
-	require.NoError(t, os.Remove(path), "failed to remove symlink")
+	// Change the timestamps of the symlink itself
```
jfly, line 74: This comment mentions symlink, but I don't see any symlinks in this function?

jfly (commented):
Thanks for working on this!

I don't think I totally understand the hashing mechanism, which I've left some ineline comments on. I'd like to understand it before I give you a final shipit.

I think it's possible that I'm confused because I'm not thinking about the scenario where we have multiple formatters interested in formatting the same file.

format/composite.go
brianmcgee: It's a correctness thing. Consider a formatter's config: 

- if `Command` changes, the outcome of applying the formatter might differ. 
- Similarly, if the underlying executable that `Command` refers to has changed, the outcome of applying the formatter might differ. 
- if `Options` change, the outcome of applying the formatter might differ.
- if `Priority` changes, the outcome of applying the formatter in a particular sequence might be different. 
- If `Includes/Excludes` changes, there may be some new files to which we should apply the formatter.

Similarly, if the global excludes change, there may be new files to which we should apply the formatters.

We shouldn't wait until the underlying file changes before doing so; that's why we bust the cache. We could try to be more nuanced and expire only files that are affected. But the complexity of doing that versus the minimal cost of busting the cache is not worth it, in my opinion.

format/composite.go
brianmcgee: Updated

format/composite.go
brianmcgee: You make a fair point. We don't need to include excludes, both globally and at the formatter level, in the hash.

format/formatter.go
brianmcgee: Agreed, see comments above.

format/formatter_test.go
brianmcgee: Yeah, fair enough. Was being lazy :)

test/test.go
brianmcgee: This is stale and needs removed.

brianmcgee:
I appreciate the feedback. Made some changes, left some comments.

format/composite.go
zimbatm: It would be nice if it only busted the cache for the current formatter group instead of the whole cache.

zimbatm (approved):

format/composite.go
brianmcgee: This is addressed in the latest changes I just pushed :slightly_smiling_face:

format/composite.go
```diff
@@ -345,6 +263,80 @@ func (c *CompositeFormatter) Close(ctx context.Context) error {
...
+// This can be used to determine if there has been a material change in config or setup that requires the cache
+// to be invalidated.
+func (c *CompositeFormatter) Hash() (string, error) {
+	h := sha256.New()
+
+	// sort formatters determinstically
+	formatters := make([]*Formatter, 0, len(c.formatters))
+	for _, f := range c.formatters {
+		formatters = append(formatters, f)
+	}
```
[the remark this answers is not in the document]
brianmcgee: `c.formatters` is a `map[string]*Formatter`.

format/composite.go
```diff
@@ -345,6 +263,80 @@ func (c *CompositeFormatter) Close(ctx context.Context) error {
 	return nil
 }
 
+// Hash takes anything that might affect the paths to be traversed, or how they are traversed, and adds it to a sha256
+// hash.
+// This can be used to determine if there has been a material change in config or setup that requires the cache
+// to be invalidated.
+func (c *CompositeFormatter) Hash() (string, error) {
+	h := sha256.New()
```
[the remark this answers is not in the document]
brianmcgee: Switched to `md5`.

brianmcgee:
Still chasing down a caching bug before I mark this ready for review again.

zimbatm (approved):
LGTM

jfly (approved):
Looks great! I'm a bit confused about how we're handling the db schema change. What happens when we run this new code with an old cache? Does this work because we've change the name of the db? (I saw a change from Sha to Sha256 when computing the filename).

brianmcgee:
> Looks great! I'm a bit confused about how we're handling the db schema change. What happens when we run this new code with an old cache? Does this work because we've change the name of the db? (I saw a change from Sha to Sha256 when computing the filename).

The first time someone runs this, it will create a new db (fresh cache) rather than use an existing one. 
There is no need to migrate the state in there. Yes, they lose their cache state, but it's not a big deal. They'll only really notice if they're running it on a huge repo.

format/composite.go
```diff
@@ -0,0 +1,193 @@
...
+		if err := file.Release(releaseCtx); err != nil {
+			return fmt.Errorf("failed to release file: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// signature takes anything that might affect the paths to be traversed,
+// or how they are traversed, and adds it to a sha256 hash.
```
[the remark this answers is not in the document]
brianmcgee: Yeah I'll clean it up.

format/scheduler.go
```diff
@@ -0,0 +1,260 @@
...
+	for _, f := range formatters {
+		if err := f.Hash(h); err != nil {
+			return nil, fmt.Errorf("failed to hash formatter %s: %w", f.Name(), err)
+		}
+	}
+
+	sig = h.Sum(nil)
+
+	// store we don't have to re-compute for each file
+	s.signatures[key] = sig
```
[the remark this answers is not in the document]
brianmcgee: Yeah I'll fix it

walk/walk.go
```diff
@@ -35,12 +34,59 @@ type File struct {
...
-	// Cache is the latest entry found for this file, if one exists.
-	Cache *cache.Entry
+	// FormattedInfo is the result of os.stat after formatting the file.
+	FormattedInfo fs.FileInfo
+
+	// FormattersSignature represents the sequence of formatters and their config that was applied to this file.
+	FormattersSignature []byte
+
+	// CachedFormatSignature is the last FormatSignature generated for this file, retrieved from the cache.
+	CachedFormatSignature []byte
```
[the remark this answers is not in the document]
brianmcgee: If there was no entry in the cache this will be `nil`.

`nil` is analogous to an empty `[]byte` for operations like `len(CacheFormatSignature` or more importantly:

```go
// compare signature with last cache entry
if bytes.Equal(signature, file.CachedFormatSignature) {
	// If the signature is the same as the last cache entry, there is nothing to do.
	// We know from the hash signature that we have already applied this sequence of formatters (and their config) to
	// this file.
	// When we applied the formatters, the file had the same mod time and file size.
	return false, nil
}
```

cmd/root_test.go
```diff
@@ -1274,15 +1306,20 @@ func treefmt(t *testing.T, args ...string) ([]byte, *stats.Stats, error) {
...
+	failOnChange := os.Getenv("TREEFMT_FAIL_ON_CHANGE") == "true" ||
+		slices.Index(args, "--fail-on-change") != -1
 
-	defer func() {
-		// Wait until we tick over into the next second before continuing.
-		// This ensures we correctly detect changes within tests as treefmt compares modtime at second level precision.
-		waitUntil := start.Truncate(time.Second).Add(time.Second)
-		time.Sleep(time.Until(waitUntil))
-	}()
+	if failOnChange {
```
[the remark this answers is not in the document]
brianmcgee: I have improved the fail-on-change test to no longer rely on this fudge.

brianmcgee:
> What's the mechanism that causes us to create a new db rather than try to use the existing one?

Without this change, we will need to add code to remove the old `formatters` bucket, and clear the `paths` bucket as it's no longer binary compatible. We could use a different bucket name other than `paths`, but we would still need to delete that bucket if it's present. 

Rather than complicating things, we can circumvent the problem by changing the scheme for determining the database name and keeping the DB code simple. 

I want to avoid a complicated migration approach for a throwaway database that will likely be discarded at some point in the near-future due to a config change or a user running `treefmt` with `-c`.

brianmcgee:
Having thought on this a bit more, we could re-use the same db if we:

- just left the old `formatters` bucket alone, it's not doing any harm and isn't that big
- deleted the old `paths` bucket
- used a new `paths_v2` bucket 

It's not _that_ onerous. It's still less code to change the naming convention for the db. It sits in `~/.cache/treefmt/eval-cache`, and it's not like anyone uses it directly.

brianmcgee:
I've created https://github.com/numtide/treefmt/issues/459 to follow up and make the tests clearer.

jfly:
> we can circumvent the problem by changing the scheme for determining the database name and keeping the DB code simple.

That makes sense! I just wasn't sure how we were ensuring we create a new db. What happens with the old db? Does it just get orphaned until someone clears their cache directory?

brianmcgee:
> What happens with the old db? Does it just get orphaned until someone clears their cache directory?

Yeah, that's right. Currently, we don't manage the cache directory e.g. clean up old dbs.

PR #463, opened 2024-10-21
feat/improve testing

### LINKED ISSUE #459: Cleanup how we test for changes
In the main `root_test.go` the idea of changes is unclear without some context. The fail on change test is particularly unclear. 

The whole thing needs a bit of a refactor,.making it clear the conditions under which changes should be detected and why.

brianmcgee:
I've gone through `root_test.go` doing some clean up and refinements to how we test.

One highlight is the introduction of a better `treefmt` helper which allows us to do things like this:

```go
treefmt(t,
	withArgs("elm/elm.json", "haskell/Nested/Foo.hs"),
	withNoError(t),
	withStats(t, map[stats.Type]int{
		stats.Traversed: 2,
		stats.Matched:   2,
		stats.Formatted: 0,
		stats.Changed:   0,
	}),
)
```

Another is using the `test-fmt-append` formatter more to generate changes and clarify how the caching behaviour should work in some tests. 

Closes #459

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -139,18 +139,31 @@ func TestCpuProfile(t *testing.T) {
 		as.NoError(os.Chdir(cwd))
 	})
 
-	_, _, err = treefmt(t, "-C", tempDir, "--allow-missing-formatter", "--cpu-profile", "cpu.pprof")
-	as.NoError(err)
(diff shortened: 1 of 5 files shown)


zimbatm (approved):
amazing

PR #469, opened 2024-11-06
ci: run golangci-lint as a flake check instead of github workflow

brianmcgee:


nix/packages/treefmt/default.nix
```diff
@@ -63,6 +63,19 @@ in
       git config --global user.name "Treefmt Test"
     '';
 
+    passthru.tests = {
+      golangci-lint = perSystem.self.treefmt.overrideAttrs (old: {
+        nativeBuildInputs = old.nativeBuildInputs ++ [pkgs.golangci-lint];
```
zimbatm, line 68: There should be a better way to say: please run this derivation with all the dependencies of the other one. But I can't think of a better approach.

zimbatm (approved):

nix/packages/treefmt/default.nix
brianmcgee: Yeah, this is the simplest approach for now.

PR #470, opened 2024-11-06
fix: unmatched reporting global excludes

### LINKED ISSUE #468: Regression on main: do not warn about missing formatters for paths in the global excludes
See https://github.com/numtide/treefmt/issues/317#issuecomment-2455111022

brianmcgee:
Improved the test which should have caught this regression and fixed the logic in composite formatter.

Closes #468


ThoFrank:
Seems to work for me. Thanks!

zimbatm (approved):

PR #471, opened 2024-11-06
fix: use PRJ_ROOT to find config

brianmcgee:
Improves the tests for config file resolution.

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -107,12 +107,26 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
 		configFile = os.Getenv("TREEFMT_CONFIG")
 	}
 
-	// find the config file if one was not specified
-	if configFile == "" {
-		if configFile, _, err = config.FindUp(workingDir, "treefmt.toml", ".treefmt.toml"); err != nil {
-			return fmt.Errorf("failed to find treefmt config file: %w", err)
-		}
+	filenames := []string{"treefmt.toml", ".treefmt.toml"}
+
+	// look in PRJ_ROOT if set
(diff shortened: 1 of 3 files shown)


cmd/root.go
```diff
@@ -107,6 +107,11 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
 		configFile = os.Getenv("TREEFMT_CONFIG")
 	}
 
+	// fallback to PRJ_ROOT
+	if configFile == "" {
+		configFile = os.Getenv("PRJ_ROOT")
```
zimbatm, line 112: If $PRJ_ROOT is set, it should look for either $PRJ_ROOT/treefmt.toml or $PRJ_ROOT/.treefmt.toml. And error out if none of those are found.

cmd/root.go
```diff
@@ -109,7 +109,13 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
...
 	// find the config file if one was not specified
 	if configFile == "" {
-		if configFile, _, err = config.FindUp(workingDir, "treefmt.toml", ".treefmt.toml"); err != nil {
+		// look in PRJ_ROOT if set, otherwise fallback to the working directory
+		searchDir := os.Getenv("PRJ_ROOT")
+		if searchDir == "" {
+			searchDir = workingDir
+		}
+
+		if configFile, _, err = config.FindUp(searchDir, "treefmt.toml", ".treefmt.toml"); err != nil {
```
zimbatm, line 118: I wonder if, if $PRJ_ROOT is set, the user intends to search from that directory or only look in that directory.

brianmcgee: Fair point. I've updated the logic and improved the test to check for this.

zimbatm (approved):

PR #480, opened 2024-11-13
README: new format

zimbatm:
Adjust the format a bit to be consistent with other projects


brianmcgee:
We can probably remove the section about moving from Rust to Go by now as well.

PR #483, opened 2024-11-16
Enable more golangci-linter

Mic92:
Fixes https://github.com/numtide/treefmt/issues/484

### DIFF
--- a/cmd/format/format.go
+++ b/cmd/format/format.go
@@ -136,7 +136,7 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 
 		relativePath, err := filepath.Rel(cfg.TreeRoot, absolutePath)
 		if err != nil {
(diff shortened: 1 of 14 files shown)


cmd/format/format.go
```diff
@@ -204,7 +204,7 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 
 	if formatErr != nil {
 		// return an error if any formatting failures were detected
-		return formatErr
+		return formatErr //nolint:wrapcheck
```
Mic92, line 207: Left out because I was unsure if this should append something to the error message.

cmd/root.go
```diff
@@ -149,5 +154,5 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
 	}
 
 	// format
-	return format.Run(v, statz, cmd, args)
+	return format.Run(v, statz, cmd, args) //nolint:wrapcheck
```
Mic92, line 157: Same as above.

brianmcgee (approved):

PR #489, opened 2024-11-29
fixup! fix: non-ascii paths in git walker

brianmcgee:


jfly (commented):
I'm afraid I can't tell what was wrong before or how this fixed it. But thanks for adding a test!

PR #490, opened 2024-11-29
fix/exit on formatter errors

brianmcgee:
Subtle bug where the formatError flag was getting reset.

## TODO
- [x] Merge #489 
- ~~[ ] Add a test to catch the original issue~~


brianmcgee:
It'll be easier to write a test for this when #458 is implemented.

jfly (commented):
Eep parallelism

PR #491, opened 2024-12-01
better error if command cannot be found

Mic92:
In this case the command was empty:

```
       > Error: failed to create composite formatter: failed to initialise formatter mypy-: formatter command not found in PATH
```


jfly (approved):
Code LGTM! The test failure appears to be unrelated. Do you have time to look into it?

PR #498, opened 2024-12-17
docs: change tagline

zimbatm:
Makes the project's purpose a bit more explicit

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,13 +21,15 @@ _A <a href="https://numtide.com/">numtide</a> project._
 
 ## Motivation
 
-It’s common to format code according to the project’s standards before making contributions to any project. This task
-seems trivial at first sight — you can set up the required language formatter in your IDE.
+Modern code repositories are rarely written in a single language. They often contain a mix of languages, each with its own formatting requirements.
 
-However, contributing to multiple projects requires more effort: you must change your formatter configurations each
-time you switch between projects or call them manually.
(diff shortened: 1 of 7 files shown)


README.md
```diff
@@ -21,13 +21,15 @@ _A <a href="https://numtide.com/">numtide</a> project._
...
+Modern code repositories are rarely written in a single language. They often contain a mix of languages, each with its own formatting requirements.
 
-However, contributing to multiple projects requires more effort: you must change your formatter configurations each
-time you switch between projects or call them manually.
+While working for our customers, we noticed that projects each tended to re-implement the same formatter multiplexing logic. A single script that invokes all the formatters.
 
-Formatting requires less effort with a universal formatter that supports multiple languages but is still project-specific.
+What if that script was a single command?
+
+What if that single command would handle all the formatters in parallel, and only format the files that have changed?
```
jfly, line 30: (nit, feel free to ignore)

> only format the files that have changed

To me, this sounds like it's talking about gradual adoption in a large repo (aka: https://github.com/numtide/treefmt/issues/311).

I think this is actually talking about speed. How about we try to make that clearer? I'm thinking something like: "..., and only spend time formatting the files that changed since you last ran it?"

jfly (approved):
I like this rebranding =)

README.md
zimbatm: thanks

PR #504, opened 2025-01-03
Verbosity changes

### LINKED ISSUE #501: A `--quiet` flag that will prevent this tool from outputing anything
Currently, regardless of the verbosity I specify, I always get at minimum:

```log
2024/12/19 18:43:16 INFO using config file: /path/to/my-repo
traversed 89 files
emitted 3 files for processing
formatted 3 files (0 changed) in 173ms
```

When in quiet mode, I don't want anything on `stdout` and only want errors on `stderr`.

brianmcgee:
I've made some changes to how verbose we are being:

- **feat: log the config file used at DEBUG level instead of INFO**
- **feat: print stats to stderr instead of stdout**
- **feat: change default log level for unmatched to INFO instead of WARN**
- **feat: add a quiet flag to suppress all output except for errors**

Closes #501

### DIFF
--- a/cmd/format/format.go
+++ b/cmd/format/format.go
@@ -197,9 +197,9 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 		return fmt.Errorf("failed to close walker: %w", err)
 	}
 
-	// print stats to stdout, unless we are processing from stdin and therefore outputting the results to stdout
-	if !cfg.Stdin {
-		statz.Print()
+	// print stats to stderr
+	if !cfg.Quiet {
+		statz.PrintToStderr()
 	}
 
 	if formatErr != nil {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -131,7 +131,7 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
 		return fmt.Errorf("failed to find treefmt config file: %w", err)
 	}
 
-	log.Infof("using config file: %s", configFile)
+	log.Debugf("using config file: %s", configFile)
 
 	// read in the config
 	v.SetConfigFile(configFile)
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -144,13 +144,19 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
 	log.SetOutput(os.Stderr)
 	log.SetReportTimestamp(false)
 
-	switch v.GetInt("verbose") {
-	case 0:
-		log.SetLevel(log.WarnLevel)
-	case 1:
-		log.SetLevel(log.InfoLevel)
-	default:
-		log.SetLevel(log.DebugLevel)
+	if v.GetBool("quiet") {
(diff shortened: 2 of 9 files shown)


cmd/format/format.go
```diff
@@ -197,8 +197,8 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 		return fmt.Errorf("failed to close walker: %w", err)
 	}
 
-	// print stats to stdout, unless we are processing from stdin and therefore outputting the results to stdout
-	if !cfg.Stdin {
+	// print stats to stdout unless we are processing from stdin or verbosity level is lower than info (1)
+	if !cfg.Stdin && cfg.Verbose > 0 {
 		statz.Print()
```
zimbatm, line 202: It might make sense to send the stats to stderr since it's a form of reporting. Stdout is typically reserved for data that can be further processed. And treefmt is already using stdout to print formatted output on `--stdin` here. Sending the statz to stderr would allow to drop the `!cfg.Stdin` check as well.

cmd/root.go
```diff
@@ -144,15 +142,21 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
 	log.SetOutput(os.Stderr)
 	log.SetReportTimestamp(false)
 
+	// set log level based on verbosity
+	// by default we only emit errors
```
zimbatm, line 146: warning should also be emitted by default. otherwise they are not warnings.

cmd/root.go
```diff
@@ -144,15 +142,21 @@ func runE(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, args []string)
...
+		log.SetLevel(log.ErrorLevel)
 	case 1:
+		log.SetLevel(log.WarnLevel)
+	case 2:
 		log.SetLevel(log.InfoLevel)
 	default:
 		log.SetLevel(log.DebugLevel)
 	}
 
+	log.Infof("using config file: %s", configFile)
```
zimbatm, line 158: This could be a debug statement since it's only useful to debug things out.

jfly (commented):
LGTM, but I agree with the feedback that @zimbatm already gave =)

cmd/root.go
brianmcgee: Made this change

stats/stats.go
```diff
@@ -34,22 +35,27 @@ func (s *Stats) Elapsed() time.Duration {
...
 		strings.Join(components, "\n"),
 		s.Value(Traversed),
 		s.Value(Matched),
 		s.Value(Formatted),
 		s.Value(Changed),
 		s.Elapsed().Round(time.Millisecond),
-	)
+	); err != nil {
+		return fmt.Errorf("failed to write to stderr: %w", err)
+	}
```
zimbatm, line 56: I would swallow that error, just like fmt.Printf does.

Reason is that if the log is not getting outputted, it doesn't affect the logic of the program.

zimbatm (commented):
One small nit but I think it's ready. :+1:

zimbatm (approved):

Issue #509, opened 2025-01-17
Add a watch mode for automatic formatting on file save


shikanime:
**Is your feature request related to a problem? Please describe.**

Yes, currently treefmt requires manual execution to format the code.

**Describe the solution you'd like**

I propose to add a new format watch mode to treefmt. This mode would allow treefmt to automatically format the code whenever the file is saved to disk. This could be achieved by integrating with IDEs or using a file watcher to detect changes, perhaps using watchman.

**Describe alternatives you've considered**

One alternative is to manually format the code after each save. However, this is inefficient and easy to forget. Another alternative is to use an IDE extension that provides automatic formatting for treefmt. However, treefmt-nix doesn't export the config file, so implementing it in every IDE would be double work.

**Additional context**

A format watch mode would improve the developer experience by automating the formatting process and ensuring a consistent code style. It would also help reduce errors caused by forgetting to format code manually.

I believe this feature would be beneficial to many treefmt users and would be in line with the tool's goal of simplifying code formatting.

jfly:
I'm (lightly) opposed to this feature. IMO, it's something your text editor/IDE should be doing instead.

> However, treefmt-nix doesn't export the config file, so implementing it in every IDE would be double work.

Which editor do you use? Neovim now has support for this (via none-ls): https://github.com/nvimtools/none-ls.nvim/blob/main/doc/BUILTINS.md#nix_flake_fmt, and it doesn't create double work for you. Perhaps you could implement something similar for your editor?

shikanime:
I don't think I'm representative of most of the community. I regularly switch between vscode, neovim and sometimes helix. And I use nix fmt a lot, but have found the nix flake command to be particularly slow to start compared to treefmt execution itself.

jfly:
> And I use nix fmt a lot, but have found the nix flake command to be particularly slow to start compared to running treefmt itself.

I encourage you to try the nix_flake_fmt builtin for none-ls for neovim that I mentioned earlier. It immediately discovers the nix fmt entry point on startup, and caches the result. In practice, I never find that I have to wait at all for it, because I usually take a few moments after opening a file before I make a change.

As for switching between editors, if they don't already have support for `nix fmt`, I'd encourage you to look into building something like that. The nix community would benefit from it, and it wouldn't even be a treefmt specific thing you'd be building.

brianmcgee:
> And I use nix fmt a lot, but have found the nix flake command to be particularly slow to start compared to running treefmt

[...]

PR #510, opened 2025-01-18
Add notify based formatter

shikanime:
Close: #509

### DIFF
--- a/walk/watch.go
+++ b/walk/watch.go
@@ -1,0 +1,149 @@
+package walk
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"github.com/fsnotify/fsnotify"
+	"github.com/numtide/treefmt/v2/stats"
+	"golang.org/x/sync/errgroup"
+)
+
+type WatchReader struct {
+	root string
+	path string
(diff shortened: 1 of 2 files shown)


walk/watch.go
```diff
@@ -0,0 +1,153 @@
...
+	// we also clean the path up in case there are any ../../ components etc.
+	fqPath := filepath.Clean(filepath.Join(root, path))
+
+	// ensure the path is within the root
+	if !strings.HasPrefix(fqPath, root) {
+		return nil, fmt.Errorf("path '%s' is outside of the root '%s'", fqPath, root)
+	}
+
+	// start watching the path
+	if err := watcher.Add(fqPath); err != nil {
```
jfly, line 148: When does this run? Is it exactly once (per path) on startup? What about the scenario where new files are created after we register all these watches?

shikanime: It will run in async mode as soon as we call the Add method, for the moment I've kept it simple so that this only formats file changes and not the initial formatting, but later this would be the preferred behaviour.

jfly: Sorry, I don't think I understand: with the code in this PR, will we start watching files if they're created after the watch starts, or would someone have to restart the watch process to watch those new files?

PR #513, opened 2025-01-23
Enforce formatter name doesn't contain bad characters

brianmcgee:
As the title suggests. 

Recently discovered in https://github.com/numtide/treefmt-nix that names with `/` could be passed through and would break execution. This enforces simple formatted names in config and when passing as `formatters` arg. 

Also fixes misuse of the assertions API.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -332,6 +332,44 @@ func TestSpecifyingFormatters(t *testing.T) {
 			}),
 		)
 	})
+
+	t.Run("bad names", func(t *testing.T) {
+		for _, name := range []string{"foo$", "/bar", "baz%"} {
+			treefmt(t,
+				withArgs("--formatters", name),
+				withError(func(as *require.Assertions, err error) {
+					as.ErrorContains(err, fmt.Sprintf("formatter name %q is invalid", name))
+				}),
+			)
+
+			t.Setenv("TREEFMT_FORMATTERS", name)
+
+			treefmt(t,
+				withError(func(as *require.Assertions, err error) {
+					as.ErrorContains(err, fmt.Sprintf("formatter name %q is invalid", name))
+				}),
+			)
+
+			t.Setenv("TREEFMT_FORMATTERS", "")
+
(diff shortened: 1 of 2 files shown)


cmd/root_test.go
```diff
@@ -331,11 +327,50 @@ func TestSpecifyingFormatters(t *testing.T) {
...
 		treefmt(t,
-			withError(func(err error) {
-				as.Errorf(err, "formatter not found in config: bar")
+			withError(func(as *require.Assertions, err error) {
+				as.ErrorContains(err, "formatter bar not found in config")
 			}),
 		)
 	})
+
+	t.Run("bad names", func(t *testing.T) {
```
jfly, line 336: Nit, feel free to ignore: It seems to me that this is testing 2 things:

1. That all forms of configuration (cli flag, env var, conf file) all work.
2. That we enforce some rules on names.

We should already have tests that confirm 1). I'd argue that testing it again is making this test more verbose than it needs to be, which distracts from the main point of the test.

config/config.go
```diff
@@ -210,12 +211,33 @@ func FromViper(v *viper.Viper) (*Config, error) {
 		cfg.Excludes = cfg.Global.Excludes
 	}
 
+	// validate formatter names do not contain invalid characters
+
+	nameRegex := regexp.MustCompile("^[a-zA-Z0-9_-]+$")
```
jfly, line 216: I can never remember: does this do the right thing with newlines? In python, ^ and $ sometimes (I think in multiline mode?) mean "beginning of line" and "end of line", which means people can sneak newline characters in. That's what `\A` and `\B` or python's newer "no really, search for a match for the *whole* string, not a part of it" function is for.

Sorry for the python esoterica. No idea if any of this applies here. Maybe just expand the test above to have a name that is actually 2 valid names separated by a newline char. Hopefully this code already does the right thing.

jfly (approved):

config/config.go
brianmcgee: Golang doesn't have an explicit multiline mode for regexes; you must explicitly add matches for new lines in your expressions. We aren't doing that.

PR #527, opened 2025-02-25
feat: remove gomod2nix

brianmcgee:
- simpler package build with closer alignment to nixpkgs
- simplifies renovate workflow
- pins go version to 1.23


jfly (approved):
One question, but LGTM.

Issue #495, opened 2024-12-10
Don't generate warnings on missing formatters for symlinks to directories


Mic92:
Treefmt seems to use lstat instead of stat to figure out a filetype?
It leads to some odd behavior when symlinks to directories are used.
Treefmt should not generate warnings for directories.

```
~/.homesick/repos/dotfiles test-foo* ⇡
nix-shell-env % nix fmt
2024/12/10 06:14:01 INFO using config file: /nix/store/9skqqjkkajkf44rg1ani1l0bcyr75cq1-treefmt.toml
WARN no formatter for path: home/.config/nixpkgs
WARN no formatter for path: home/.fast-syntax-highlighting
WARN no formatter for path: home/.gef-extras
WARN no formatter for path: home/.gef-mic92
traversed 999 files
emitted 346 files for processing
formatted 1 files (0 changed) in 31ms
```

```
~/.homesick/repos/dotfiles test-foo* ⇡
nix-shell-env % ls -la home/.gef-*
lrwxrwxrwx joerg users 17 B 5 months ago  home/.gef-extras@ ⇒ ../gdb/gef-extras
lrwxrwxrwx joerg users 16 B 5 months ago  home/.gef-mic92@ ⇒ ../gdb/gef-mic92
```

```
nix-shell-env % stat home/.gef-extras
  File: home/.gef-extras -> ../gdb/gef-extras
  Size: 17              Blocks: 1          IO Block: 512    symbolic link
Device: 0,53    Inode: 6577745     Links: 1
Access: (0777/lrwxrwxrwx)  Uid: ( 1000/   joerg)   Gid: (  100/   users)
Access: 2024-12-09 09:00:15.636895706 +0100
Modify: 2024-07-01 16:23:33.678525570 +0200
Change: 2024-07-01 16:23:45.254346582 +0200
 Birth: 2024-07-01 16:23:33.678525570 +0200
```

```
stat ./gdb/gef-extras
  File: ./gdb/gef-extras
  Size: 19              Blocks: 17         IO Block: 1536   directory
Device: 0,53    Inode: 132444      Links: 9
Access: (0755/drwxr-xr-x)  Uid: ( 1000/   joerg)   Gid: (  100/   users)
Access: 2024-12-09 08:42:23.294971942 +0100
Modify: 2024-10-31 10:32:46.812263967 +0100
Change: 2024-10-31 10:32:46.812263967 +0100
 Birth: 2024-07-01 16:23:33.494528414 +0200
```

[...]

PR #528, opened 2025-03-03
fix: ignoring symlinks

gabyx:
@brianmcgee : I tried the following:

- Ignore symlinks on the Git reader.
- Adding a symlink to the examples, to test if it gets ignored everywhere this examples are used.
- Add resolving symlinks on the `Run` functions to pass the resolved paths to the readers, such that the readers `Filesystem` or `Git` can ignore symlinks always. `treefmt symlink.yaml`  or `treefmt /a-symlink/b/c/d.yaml` should be allowed.

See some comments I need help on which is unclear to me.

Fixes: #495


cmd/format/format.go
```diff
@@ -88,8 +88,8 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 
 		// ensure db is closed after we're finished
 		defer func() {
-			if err := db.Close(); err != nil {
-				log.Errorf("failed to close cache: %v", err)
+			if e := db.Close(); e != nil {
```
gabyx, line 91: some shadowing lints.

cmd/format/format.go
```diff
@@ -129,27 +129,36 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
...
-		if err != nil {
-			return fmt.Errorf("error computing relative path from %s to %s: %w", cfg.TreeRoot, absolutePath, err)
+		if walkType != walk.Stdin {
+			if _, e = os.Stat(absolutePath); e != nil {
+				return fmt.Errorf("path %s not found", path)
+			}
+
+			// // symlinks are allowed on `paths` input, we resolve them here, since
+			// // the readers will k
+			// realPath, ee := filepath.EvalSymlinks(absolutePath)
```
gabyx, line 144: @brianmcgee : I tried here to test the `filepath.EvalSymlinks` (and then remove the `Stat` because `EvalSymlinks` should do that already) to resolve symlinks but 
the function does error out in tests? I dont get it: It does not find the file?

gabyx:
@brianmcgee : I have question to the test here:

Why do we even test to run in the `.git` folder? You will never want that IMO, or is this really feature?
Shouldn't treefmt relaly ignore any `.git` directory (got from `git rev-parse --git-common-dir` in the tree root)

https://github.com/numtide/treefmt/blob/bdd8687ef1e83868ed3614b77301f8aa63b4c3ef/cmd/root_test.go#L1401

brianmcgee:
> Why do we even test to run in the `.git` folder? You will never want that IMO, or is this really a feature? Shouldn't treefmt ignore any `.git` directory (got from `git rev-parse --git-common-dir` in the tree root)

In that particular test, it's checking that if we momentarily switch to the filesystem walker, the `.git` directory gets picked up and traversed. It's not that we want to traverse the `.git` directory per se, but that it should traverse any directory when the filesystem walker is enabled.

gabyx:
> > Why do we even test to run in the `.git` folder? You will never want that IMO, or is this really a feature? Shouldn't treefmt ignore any `.git` directory (got from `git rev-parse --git-common-dir` in the tree root)
> 
> 
> 
> In that particular test, it's checking that if we momentarily switch to the filesystem walker, the `.git` directory gets picked up and traversed. It's not that we want to traverse the `.git` directory per se, but that it should traverse any directory when the filesystem walker is enabled. 
>
Ah ok.

I need to still refactor the symlink resolving code in the Run function, its not good yet. I think the only correct way is to us EvalSymlinks and check for errors and make the test pass. As this is only done on path inputs it should not degrade performance.

gabyx:
@brianmcgee: The thins is done. Happy for a review what you think about the behavior. Hope its ok like that.

brianmcgee:
Thanks for this, I'll try and review it in the next few days.

brianmcgee:
@gabyx please rebase on latest `main`

cmd/format/format.go
```diff
@@ -237,3 +216,40 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
...
+
+		if strings.HasPrefix(relativePath, "..") {
+			return fmt.Errorf("path %s not inside the tree root %s", path, treeRoot)
+		}
+
+		paths[i] = relativePath
+	}
+
+	return nil
+}
```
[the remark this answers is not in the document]
brianmcgee: If `filepath.EvalSymlinks` returns an error can we safely assume it's always because the path was not found? 

You should be able to perform an explicit check with `errors.Is(os.ErrNotExist)` for the path not found scenario, and a generic `else if err != nil` to account for the general case, mixing in the error into the `fmt.Errorf` so we can see exactly what it was.

brianmcgee (changes requested):

gabyx:
@brianmcgee: Did the changes and rebased.

brianmcgee:
Thanks :pray: I'll give it another review in the next day or so

brianmcgee (approved):

Issue #529, opened 2025-03-03
Add static as well as dynamic shell completion scripts


a-kenji:
**Is your feature request related to a problem? Please describe.**

When using `treefmt` I am always trying to autocomplete flags.

Especially when setting `treefmt --formatters [FORMATTER..]`.

**Describe the solution you'd like**

I would like to have static completions for all the flags that treefmt provides.

I would also really like dynamic completions, especially for the `--formatters` flags, since I don't always want to look into the configuration how the formatter is actually named right now.
I expect this could help save some time and frustration for people.

**Describe alternatives you've considered**

**Additional context**

brianmcgee:
It would appear [cobra] already added this for us; it was just hidden and [poorly documented]:

```console
❯ result/bin/treefmt completion
Generate the autocompletion script for treefmt for the specified shell.
See each sub-command's help for details on how to use the generated script.

Usage:
  treefmt completion [command]

Available Commands:
  bash        Generate the autocompletion script for bash
  fish        Generate the autocompletion script for fish
  powershell  Generate the autocompletion script for powershell
  zsh         Generate the autocompletion script for zsh

Flags:
  -h, --help   help for completion

Use "treefmt completion [command] --help" for more information about a command.
```

@a-kenji give it a go and let me know how you get on. 

I think this isn't showing up in the help because of how we take over the root command. I've been wanting to switch to a sub command approach for a while, with `format` being the default sub command. This would be the first breaking change in the cli. 

In the meantime, if this works well, we can update the docs at least to make it clearer.

[cobra]: https://cobra.dev/
[poorly documented]: https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md

zimbatm:
It currently doesn't work because it conflicts with the treefmt root command. It thinks that the user wants to format the "./completion" sub-folder.

```diff
diff --git i/pkgs/by-name/tr/treefmt2/package.nix w/pkgs/by-name/tr/treefmt2/package.nix
index 0aca91e20c5c..f9a4b98d59d7 100644
--- i/pkgs/by-name/tr/treefmt2/package.nix
+++ w/pkgs/by-name/tr/treefmt2/package.nix
@@ -2,6 +2,7 @@
   lib,
   buildGoModule,
   fetchFromGitHub,
+  installShellFiles,
 }:
 buildGoModule rec {
   pname = "treefmt";
@@ -20,6 +21,8 @@ buildGoModule rec {
 
[...]

PR #542, opened 2025-03-11
fix: shell completions

brianmcgee:
- disables default `completion` cobra sub-command
- adds a `--completion [bash|zsh|fish]` flag which will generate shell completions using cobra's built-in support
- adds `installShellCompletion` to the `treefmt` package (this will need added to the nixpkgs entry after release)

The completions are static and address the first half of #529.

### DIFF
--- a/cmd/completions.go
+++ b/cmd/completions.go
@@ -1,0 +1,29 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/spf13/cobra"
+)
+
+func generateShellCompletions(cmd *cobra.Command, args []string) error {
+	var err error
+
+	switch args[0] {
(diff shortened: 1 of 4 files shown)


brianmcgee:
@a-kenji This fixes the default static completion behaviour that `cobra` provides for "free".

Can you give it a poke and see how it behaves? 

I've never implemented completions, so some guidance and help testing would be appreciated. 

When we get these static completions working, we can have a look at [dynamic completions](https://github.com/spf13/cobra/blob/main/site/content/completions/_index.md#dynamic-completion-of-nouns).

a-kenji:
I tested it and it behaves how I would assume it would.
Thank you, this is already a much better experience.

PR #551, opened 2025-03-13
chore(build): check for errors when running command

jfly:
`os.system` happily returns an exit code if the command you're running fails:

```
$ python
>>> import os
>>> os.system("echo uh oh && exit 42"); print("This happened")
uh oh
10752
This happened
```

### DIFF
--- a/docs/hooks.py
+++ b/docs/hooks.py
@@ -4,7 +4,7 @@ from pathlib import Path
 def on_pre_build(**kwargs):
     with Path("./snippets/usage.txt").open("w") as f:
         subprocess.run(
-            ["nix", "run", ".#treefmt", "--", "--help"],
+            ["treefmt", "--help"],
             stdout=f,
             check=True,
         )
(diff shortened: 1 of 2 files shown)


docs/hooks.py
```diff
@@ -1,4 +1,10 @@
...
 
 def on_pre_build(**kwargs):
-    os.system('nix run .#treefmt -- --help > ./snippets/usage.txt')
\ No newline at end of file
+    with Path("./snippets/usage.txt").open("w") as f:
+        subprocess.run(
+            ["nix", "run", ".#treefmt", "--", "--help"],
+            stdout=f,
+            check=True,
+        )
```
jfly, line 10: A much simpler alternative:

```suggestion
    assert 0 == os.system('nix run .#treefmt -- --help > ./snippets/usage.txt')
```

I personally like to avoid bash/sh/whatever if I can, though. One less dependency.

brianmcgee: I changed this to executing the `treefmt` binary directly, as this causes issues in the `docs` package https://github.com/numtide/treefmt/pull/551/commits/583681d0794690c2544bf65db057245e50a130fb

brianmcgee (approved):

Issue #549, opened 2025-03-13
Define treefmt package in `callPackage` style


brianmcgee:
This should help reduce the impedance mismatch between how we define the package in the flake and how we define the package in `nixpkgs`. 

Ideally we can just copy-paste when we do a release.

PR #554, opened 2025-03-13
chore: define `treefmt` package in `callPackage` style

jfly:
This completes https://github.com/numtide/treefmt/issues/549

### DIFF
--- a/nix/packages/treefmt/package.nix
+++ b/nix/packages/treefmt/package.nix
@@ -1,0 +1,49 @@
+{
+  buildGoModule,
+  installShellFiles,
+  lib,
+  pname,
+  src,
+  version,
+  passthru,
+}:
+buildGoModule rec {
+  inherit pname version src passthru;
+
+  vendorHash = "sha256-A8Gw3kEpw8iYBMw6GSnJdJAaIhZp+zMOoBK4eWwo8lU=";
+
+  subPackages = ["."];
+
+  env.CGO_ENABLED = 0;
+
+  nativeBuildInputs = [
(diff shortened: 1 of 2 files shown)


nix/packages/treefmt/package.nix
```diff
@@ -0,0 +1,49 @@
+{
+  buildGoModule,
+  installShellFiles,
+  lib,
+  pname,
+  src,
+  version,
+  passthru,
```
jfly, line 8: This was the funkiest part of this refactor. We seem to be using `passthru` for 2 things:

1. Stash a coverage test somewhere so `.github/workflows/coverage.yml` can invoke it with `nix build .#treefmt.tests.coverage`.
2. Stash the update vendor hash utility somewhere so `.github/workflows/update-vendor-hash.yml` can invoke it with `nix run .#treefmt.update-vendor-hash`.

To minimize our diff with nixpkgs, I didn't want leave these in `package.nix`, so I opted to pass them through like this. I'm very open to alternate ideas for how to tackle this, though.

brianmcgee:
The original idea behind this PR was to reduce the impedance mismatch between the flake and nixpkgs packages and make it easier to copy-paste when releasing to nixpkgs. 

With the introduction of https://github.com/NixOS/nixpkgs/commit/447f275222a911e7ec2ce3f1741b00198eb1f0ae, I think this will become more difficult, and I see less value in trying to achieve it now. 

Probably best to implement each in their respective idiomatic approaches and deal with the minor amount of adapting one to the other when we release.

PR #560, opened 2025-03-21
feat: traverse untracked files in git walker

### LINKED ISSUE #557: `treefmt` ignores files which are untracked but not gitignored
**Describe the bug**

If you have a file which is not yet tracked by git, but which is not gitignored, then `treefmt` does not format it (because it doesn't show up in `git ls-files`).

I think this could be fixed by passing `-o` to `git ls-files`.

**To Reproduce**

1. Add a new file
2. Don't `git add` it
3. Run `treefmt`

**Expected behavior**

I would expect it to be formatted.

brianmcgee:
This was quick enough to implement and, at face value, feels appropriate. 

But my spidey senses are tingling, and I can't quite understand why. 

@zimbatm, @jfly: why might this be the wrong thing to do?

Closes #557

### DIFF
--- a/walk/git.go
+++ b/walk/git.go
@@ -38,7 +38,7 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error) 
 		r, w := io.Pipe()
 
 		// create a command which will execute from the specified sub path within root
-		cmd := exec.Command("git", "ls-files")
+		cmd := exec.Command("git", "ls-files", "--cached", "--others")
 		cmd.Dir = filepath.Join(g.root, g.path)
 		cmd.Stdout = w
 
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1359,11 +1359,15 @@ func TestGit(t *testing.T) {
 	as.NoError(gitCmd.Run(), "failed to init git repository")
 
 	// run before adding anything to the index
+	// we should pick up untracked files since we use `git ls-files -o`
 	treefmt(t,
 		withConfig(configPath, cfg),
 		withNoError(t),
 		withStats(t, map[stats.Type]int{
-			stats.Traversed: 0,
+			stats.Traversed: 31,
+			stats.Matched:   31,
+			stats.Formatted: 31,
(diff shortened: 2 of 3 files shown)


zimbatm:
The only downside I can think of is a combination of a build system outputting files in the repo, and the developer not adding those artefacts to the .gitignore. Eg: node_modules getting formatted and creating reproducibility issues. It can be punctually annoying, but also easily fixed by adding those into the .gitignore.

walk/git.go
```diff
@@ -38,7 +38,9 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error)
 		r, w := io.Pipe()
 
 		// create a command which will execute from the specified sub path within root
-		cmd := exec.Command("git", "ls-files")
+		// we pass `-c` to pick up any files in the git index (this is the default ls-files behaviour)
+		// we pass `-o` to pick up untracked files
+		cmd := exec.Command("git", "ls-files", "-co")
```
jfly, line 43: nit: IMO, using the long form of these options (`--cached` and `--ignored`) would be slightly more readable and reduce the need for the documentation explaining the options.

brianmcgee:
It has more to do with how clean it currently is: if it's tracked, we traverse it; if it's not, we don't. If you want something else, use the `file` walker. 

That being said, I can see the benefit of this. As they say round these parts, let's suck it and see.

walk/git.go
```diff
@@ -38,7 +38,7 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error)
 		r, w := io.Pipe()
 
 		// create a command which will execute from the specified sub path within root
-		cmd := exec.Command("git", "ls-files")
+		cmd := exec.Command("git", "ls-files", "--cached", "--others")
```
zowoq, line 41: Probably want to add `--exclude-standard` so it respects gitignore.

brianmcgee: Thanks https://github.com/numtide/treefmt/pull/565

PR #571, opened 2025-04-11
feat: find tree root via a user-specified command

### LINKED ISSUE #530: Dynamic tree root detection

Sometimes the project root cannot be easily identified, for example by generic tools such as treefmt wrappers.

This came up while reviewing `pkgs.nixfmt-tree`, a wrapper tool that uses treefmt to make nixfmt easier to use in simple scenarios.

In an ideal world, we wouldn't have to do any work to identify the root of the project. However `treefmt` is used in many scenarios, so I don't think that'll ever be possible.

I think the most generic solution would be to allow configuring a script (or other executable) that can be used to test whether or not a given directory is the project root.

I propose adding a new setting: `tree-root-tool`, `tree-root-test`, `tree-root-exe` (or some other name). When defined, the specified script would be run instead of testing for `tree-root-file`.

The script could _either_ be passed the directory being tested as a CLI argument or be run with that directory as its current working directory (or both, I guess).

If the script exits 0, the directory is the root. If the script exits 1, the directory is not the root. If the script has another exit code, an error has occurred.

Perhaps it would be simpler to call the script with the starting directory and have it print the root directory to stdout. That way, such a script could be implemented using `git rev-parse --show-toplevel`. This also has the advantage that the script only needs to be invoked once.

Maybe both approaches could be supported by using multiple setting names, or by reading stdout before falling back to the exit code strategy. Although this feels a little excessive.

Alternatively, we could define `tree-root-file` or `tree-root-files` as a list of files, any of which being present would indicate the project root. This may be nice to have, but isn't as powerful as running some arbitrary executable.

Another alternative may be for `nix fmt` to support specifying the flake root, e.g. as an environment variable. This would be useful, but is out of the hands of treefmt and is only useful when treefmt is being run via `nix fmt`.

Another alternative may be to offer an setting that tells treefmt to find the root using `git rev-parse --show-toplevel` instead of looking for `tree-root-file`.

For simple tools like `nixpkgs-tree`, it'd probably also be ok to _not_ attempt to find the project root at all, and instead always treat the current working directory as the tree's root. That way we wouldn't have to try and guess at how to detect the project root. Perhaps an option to enable this behavior could be added?

**Additional context**

- https://github.com/NixOS/nixpkgs/pull/384857#discussion_r1979553258
- https://github.com/NixOS/nixpkgs/pull/384857#issuecomment-2698764313

cc @jfly @infinisil

brianmcgee:
Closes #530


config/config.go
```diff
@@ -186,19 +195,46 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+			out, cmdErr := cmd.CombinedOutput()
+			if cmdErr != nil {
+				return nil, fmt.Errorf("failed to run tree-root-cmd: %w", cmdErr)
+			}
+
+			cfg.TreeRoot = strings.TrimSpace(string(out))
+		}
+	}
+
+	if cfg.TreeRoot == "" {
```
jfly, line 235: Do we want this to happen if `tree-root-cmd` successfully prints empty string? Wouldn't that be confusing for our users?

config/config.go
```diff
@@ -104,6 +108,11 @@ func SetFlags(fs *pflag.FlagSet) {
 		"The root directory from which treefmt will start walking the filesystem (defaults to the directory "+
 			"containing the config file). (env $TREEFMT_TREE_ROOT)",
 	)
+	fs.String(
+		"tree-root-cmd", "",
+		"Command to run to find the tree root (if --tree-root or --tree-root-file is not passed). (env"+
```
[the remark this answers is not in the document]
brianmcgee: Went for the less tedious approach

config/config.go
```diff
@@ -186,19 +195,46 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
 				return nil, fmt.Errorf("failed to find tree-root based on tree-root-file: %w", err)
 			}
-		} else {
-			// otherwise fallback to the directory containing the config file
-			cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
 		}
+
+		if cfg.TreeRootCmd != "" {
+			// use the output of the tree root command as the tree root
+			parts := strings.Fields(cfg.TreeRootCmd)
```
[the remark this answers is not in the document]
brianmcgee: Brought in a port of `shlex` to handle this.

config/config.go
brianmcgee: Re-ordered the logic to account for this.

jfly: I don't see how this has changed, sorry. If I use something like `--tree-root-cmd="echo -n ''"`, won't that command successfully print `""`, and then we'll go into this `if` statement?

zimbatm (approved):
it's good

config/config.go
brianmcgee: I hallucinated a change which I didn't actually commit. Refined it a bit more there to make the logic explicity more clearly mutually exclusive.

config/config.go
```diff
@@ -186,19 +196,50 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
-			if err != nil {
-				return nil, fmt.Errorf("failed to find tree-root based on tree-root-file: %w", err)
-			}
-		} else {
-			// otherwise fallback to the directory containing the config file
-			cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
+	// NOTE: at the treefmt command level we ensure that `--tree-root`, `--tree-root-cmd` and `--tree-root-file` are
+	// mutually exclusive
+	switch {
+	case cfg.TreeRoot == "" && cfg.TreeRootFile != "":
```
jfly, line 208:
```suggestion
	case cfg.TreeRootFile != "":
```

config/config.go
```diff
@@ -186,19 +196,50 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+		// search the tree root using the --tree-root-file if specified
+		_, cfg.TreeRoot, err = FindUp(cfg.WorkingDirectory, cfg.TreeRootFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find tree-root based on tree-root-file: %w", err)
+		}
+
+	case cfg.TreeRoot != "" && cfg.TreeRootCmd != "":
```
jfly, line 215:
```suggestion
	case cfg.TreeRootCmd != "":
```

brianmcgee:
@MattSturgeon, do you mind giving this a review as well?

config/config.go
```diff
@@ -186,19 +196,50 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
-		if cfg.TreeRootFile != "" {
-			// search the tree root using the --tree-root-file if specified
-			_, cfg.TreeRoot, err = FindUp(cfg.WorkingDirectory, cfg.TreeRootFile)
-			if err != nil {
-				return nil, fmt.Errorf("failed to find tree-root based on tree-root-file: %w", err)
-			}
-		} else {
-			// otherwise fallback to the directory containing the config file
-			cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
+	// NOTE: at the treefmt command level we ensure that `--tree-root`, `--tree-root-cmd` and `--tree-root-file` are
```
jfly, line 205: Is this comment about mutual exclusivity a bit misleading? Right above this, we set `cfg.TreeRootCmd` in a way that I believe could violate the exclusivity promise.

If I'm misunderstanding and they truly are exclusive by this point, then I believe the case statements here can be simplified (I've left comments below).

brianmcgee: Still had a bug, had a bit too much chocolcate yesterday. 

Here's the whole section as it currently is, with the meat of the sections removed for brevity:

```go
	switch {
	case cfg.TreeRoot == "" && cfg.TreeRootFile != "":

	case cfg.TreeRoot == "" && cfg.TreeRootCmd != "":				

	case cfg.TreeRoot == "":
		// no tree root was specified, so we fall back to the directory containing the config file
		cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
	}
```

To start with, `--tree-root` can be passed to set `cfg.TreeRoot`. That is why we begin each section with a check that `cfg.TreeRoot == ""`. 

If `--tree-root` was not specified but `--tree-root-file` was, we attempt the file-based approach. If `--tree-root` wasn't specified and `--tree-root-cmd` was, we try the command-based approach. Both of these will trigger first if possible. 

Otherwise, if `--tree-root` wasn't specified, we try the directory where the config file is located. 

> [!IMPORTANT]
> Select statements in Go do not automatically fall through. You must explicitly force this by providing the `fallthrough` statement. You can think of above as an `if-else` statement

jfly: Sorry, but I'm still confused. If these 3 settings are mutually exclusive, then we shouldn't have to be checking multiple things: if one is specified, then we know the others are not (because of mutual exclusivity).

brianmcgee: @jfly hold off on reviewing this any further.

I want to take a fresh pass later today and see if I can make all this logic easier to follow.

config/config.go
```diff
@@ -186,19 +196,50 @@ func FromViper(v *viper.Viper) (*Config, error) {
 		cfg.Walk = walk.Stdin.String()
 	}
 
+	// set git-based tree root command if the walker is Git
+	if cfg.Walk == walk.Git.String() {
+		cfg.TreeRootCmd = "git rev-parse --show-toplevel"
+	}
```
MattSturgeon, line 202: This looks like it is overriding the user-provided `tree-root-cmd` setting when in a git tree?

config/config.go
```diff
@@ -186,19 +196,50 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+	case cfg.TreeRoot == "" && cfg.TreeRootFile != "":
+		// search the tree root using the --tree-root-file if specified
+		_, cfg.TreeRoot, err = FindUp(cfg.WorkingDirectory, cfg.TreeRootFile)
+		if err != nil {
+			return nil, fmt.Errorf("failed to find tree-root based on tree-root-file: %w", err)
+		}
+
+	case cfg.TreeRoot == "" && cfg.TreeRootCmd != "":
+		// use the output of the tree root command as the tree root
+		parts, splitErr := shlex.Split(cfg.TreeRootCmd)
```
MattSturgeon, line 217: Should the use of shlex to parse the cmd be documented somewhere?

Alternatively, would it make sense to use `bash` to evaluate the command (`bash -c "$cmd"`)?

Using a "real" shell to eval the command may be more intuitive for users; e.g. they can use things like `$PWD` or `$(pwd)` within their command string. In the future, treefmt could also set its own env vars that the command would have access to.

Another alternative would be to allow `tree-root-cmd` to be either a `str` or `str[]`. Rather than parsing `'` syntax, multiple args could be separated as list elements. I guess this is only useful in toml though, not env-vars or cli-flags.

MattSturgeon (commented):
### LGTM

The concept looks good to me, thanks! I'll defer to @jfly for the more technical review.

#### shlex

I do wonder about the merit of using shlex vs either requiring that multi-arg commands be configured as a list, or evaluated using a real shell (i.e. bash).

#### treefmt-nix

Slightly off topic, but I wonder what implications this has for treefmt-nix?

Now that `git rev-parse --show-toplevel` is used by default, does that mean that treefmt-nix no longer needs to provide a default for `tree-root-file`, [`projectRootFile`](https://flake.parts/options/treefmt-nix#opt-perSystem.treefmt.projectRootFile) (default `".git/config"`), [`projectRoot`](https://flake.parts/options/treefmt-nix#opt-perSystem.treefmt.projectRoot) (default `self`[^1]), etc?

Also, how should the change be communicate to users who have gotten in the habit of setting `tree-root-file` themselves, in a way that is now likely to be redundant?

[^1]: Tangent: does `projectRoot`'s current default `self` make sense? Since that'll be an in-store path, and won't ever be used _anyway_ because `projectRootFile` is always defined

config/config.go
brianmcgee: Good catch.

config/config.go
brianmcgee: I agree with @Mic92 on this one.

cmd/root.go
```diff
@@ -46,8 +46,12 @@ func NewRoot() (*cobra.Command, *stats.Stats) {
...
 	config.SetFlags(fs)
 
-	// xor tree-root and tree-root-file flags
-	cmd.MarkFlagsMutuallyExclusive("tree-root", "tree-root-file")
+	// xor tree-root, tree-root-cmd and tree-root-file flags
+	cmd.MarkFlagsMutuallyExclusive(
+		"tree-root",
+		"tree-root-cmd",
+		"tree-root-file",
+	)
```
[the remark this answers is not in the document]
brianmcgee: The order of precedence is well-defined and provided by viper/cobra: https://treefmt.com/latest/getting-started/configure/

Regarding the XOR, I believe it should be enforced across all three methods of setting it, but I will confirm this.

config/config.go
brianmcgee: I've re-worked the logic and caught some edge cases / incorrect assumptions.

brianmcgee:
Took another pass and fixed some bad assumptions. 

Fresh reviews have been requested :pray:

cmd/root_test.go
```diff
@@ -1531,6 +1531,109 @@ func TestGit(t *testing.T) {
...
+	permutations := [][]string{
+		{"tree-root", "tree-root-cmd"},
+		{"tree-root", "tree-root-file"},
+		{"tree-root-cmd", "tree-root-file"},
+		{"tree-root", "tree-root-cmd", "tree-root-file"},
+	}
+
+	// TODO we should also test mixing the various methods in the same test e.g. env variable and config value
+	// Given that ultimately everything is being reduced into the config object after parsing from viper, I'm fairly
+	// confident if these tests all pass then the mixed methods should yield the same result.
```
jfly, line 1592: I'm having trouble understanding this TODO. Is it saying we should make a change here, or is it saying we're pretty confident things are ok and therefore we don't need to do anything?

cmd/root_test.go
```diff
@@ -1531,6 +1531,109 @@ func TestGit(t *testing.T) {
...
+
+	assertExclusiveConfig := func(as *require.Assertions, err error) {
+		as.ErrorContains(err,
+			"only one of tree-root, tree-root-cmd or tree-root-file can be specified",
+		)
+	}
+
+	envValues := map[string][]string{
+		"tree-root":      {"TREEFMT_TREE_ROOT", "bar"},
+		"tree-root-cmd":  {"TREEFMT_TREE_ROOT_CMD", "'echo buzz'"},
```
jfly, line 1561: nit: I know this does not actually matter for this test, but this looks like a weird command.

```suggestion
			cfg.TreeRootCmd = "echo buzz"
```

(same comment throughout the test)

cmd/root_test.go
```diff
@@ -1531,6 +1531,109 @@ func TestGit(t *testing.T) {
...
+		},
+		"tree-root-cmd": func(cfg *config.Config) {
+			cfg.TreeRootCmd = "'echo buzz'"
+		},
+		"tree-root-file": func(cfg *config.Config) {
+			cfg.TreeRootFile = ".git/config"
+		},
+	}
+
+	permutations := [][]string{
```
jfly, line 1583: nit: "permutations" is not the right word here. "permutation" is about the order of things, and this is not about ordering.

Mathematically, I think it's all subsets of the 3 parameters with more than 1 element (`all_subsets - subsets_of_length_1 - subsets_of_length_0 = 2 ** 3 - 3 - 1 = 4`). That's a mouthful, though. Perhaps "conflictingArgSets"?

I'd also be fine with not actually checking all conflicts. This list would start to get pretty silly if we added a 4th parameter, for instance.

config/config.go
```diff
@@ -263,6 +258,105 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+	if count > 1 {
+		return errors.New("only one of tree-root, tree-root-cmd or tree-root-file can be specified")
+	}
+
+	// set git-based tree root command if the walker is git and no tree root has been specified
+	if cfg.Walk == walk.Git.String() && count == 0 {
+		cfg.TreeRootCmd = "git rev-parse --show-toplevel"
+
+		log.Infof(
+			"git walker enabled, but tree root has not been specified: defaulting tree-root-cmd to '%s'",
```
jfly, line 290: nit, feel free to ignore: The tone of "but" sounds to me like something unexpected or surprising has happened here, and I don't think that's the case.

```suggestion
			"git walker enabled, and tree root has not been specified: defaulting tree-root-cmd to '%s'",
```

config/config_test.go
```diff
@@ -457,6 +469,50 @@ func TestTreeRootFile(t *testing.T) {
...
+	checkValue(filepath.Join(tempDir, "bar"))
+
+	// empty tree-root-cmd value
+	// should be the same as the current working directory
+	// we switch to a new temp directory to test this
+	tempDir2 := t.TempDir()
+	t.Chdir(tempDir2)
+
+	as.NoError(flags.Set("tree-root-cmd", "echo ''"))
+	checkValue(tempDir2)
```
jfly, line 513: I personally find this to be confusing behavior. If somebody writes a script for a `tree-root-cmd` and accidentally forgets to print something, wouldn't it be better for them to get an error, rather than having treefmt fallback to some undocumented behavior?

jfly (commented):
I find the new `determineTreeRoot` function to be way more readable than before, thank you!

cmd/root_test.go
MattSturgeon: > I'm fairly confident if these tests all pass then the mixed methods should yield the same result.

To me this says the current tests passing yields high confidence, but additional tests would be beneficial for completeness.

cmd/root_test.go
MattSturgeon: I think its deliberate that the `TREEFMT_TREE_ROOT_CMD` env var is being used here, rather than directly mutating the `cfg` struct. Is it the nested quotes that you are primarily complaining about?

config/config_test.go
MattSturgeon: Agreed, this should either be clearly documented or throw a user-facing error. An error is my preference.

config/config.go
MattSturgeon: We can also drop the comma if switching to "and":

```suggestion
			"git walker enabled and tree root has not been specified: defaulting tree-root-cmd to '%s'",
```

cmd/root_test.go
brianmcgee: > To me this says the current tests passing yields high confidence, but additional tests would be beneficial for completeness.

^This

cmd/root_test.go
brianmcgee: As @MattSturgeon says. However it doesn't technically require the `' '` when used in an env variable, that's just a copy paste error.

brianmcgee: I changed the example and removed the `' '` in the env variable examples.

cmd/root_test.go
brianmcgee: I've gone with `invalidCombinations`.

config/config_test.go
brianmcgee: I have changed it to throw an error if the `tree-root-cmd` returns an empty output.

brianmcgee:
> #### treefmt-nix
> Slightly off topic, but I wonder what implications this has for treefmt-nix?
> 
> Now that `git rev-parse --show-toplevel` is used by default, does that mean that treefmt-nix no longer needs to provide a default for `tree-root-file`, [`projectRootFile`](https://flake.parts/options/treefmt-nix#opt-perSystem.treefmt.projectRootFile) (default `".git/config"`), [`projectRoot`](https://flake.parts/options/treefmt-nix#opt-perSystem.treefmt.projectRoot) (default `self`[1](#user-content-fn-1-ef9675651c9ae9cc6b3e383203b222a3)), etc?

Correct. 

The default root file of `.git/config` can be removed. I'll apply that change when I next update `treefmt-nix` to incorporate this change (after the release).
 
> Also, how should the change be communicate to users who have gotten in the habit of setting `tree-root-file` or `projectRootFile` themselves, as that is now likely to be redundant?

This is a good point. It will be in the release notes, and I can point it out on Fosstodon as well. 

A Nixos discourse post might also be appropriate?

config/config.go
```diff
@@ -263,6 +258,111 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+	// execute
+	out, cmdErr := cmd.CombinedOutput()
+	if cmdErr != nil {
+		log.Errorf("tree-root-cmd output: \n%s", out)
+
+		return "", fmt.Errorf("failed to run tree-root-cmd: %w", cmdErr)
+	}
+
+	// trim the output and check it's not empty
```
MattSturgeon, line 356: Should we also check the output is not multi-line? i.e. has no `"\n"` infix.

And/or check that the output is a real directory - or I guess that is handled later?

config/config.go
```diff
@@ -263,6 +258,105 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+		if cfg.TreeRoot, err = execTreeRootCmd(cfg); err != nil {
+			return err
+		}
+
+	default:
+		// no tree root was specified
+		log.Debugf("no tree root specified, defaulting to the directory containing the config file: %s", v.ConfigFileUsed())
+
+		cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
```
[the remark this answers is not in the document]
MattSturgeon: I guess https://treefmt.com/latest/getting-started/configure/ answers my question here, the wording clearly implies there must always be a config file.

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/adrg/xdg v0.5.3
 	github.com/charmbracelet/log v0.4.1
 	github.com/gobwas/glob v0.2.3
+	github.com/google/shlex v0.0.0-20191202100458-e7afc7fbc510
```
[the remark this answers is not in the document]
brianmcgee: I came to the same conclusion, in that it doesn't need active maintenance so I'm happy to lean on it.

config/config.go
```diff
@@ -263,6 +258,105 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+	// set a reasonable timeout of 2 seconds to wait for the command to return
+	// it shouldn't take anywhere near this amount of time unless there's a problem
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
+
+	// construct the command, setting the correct working directory
+	//nolint:gosec
+	cmd := exec.CommandContext(ctx, parts[0], parts[1:]...)
+	cmd.Dir = cfg.WorkingDirectory
```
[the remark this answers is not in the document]
brianmcgee: Can't hurt to document it.

config/config.go
```diff
@@ -263,6 +258,111 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+	// execute
+	out, cmdErr := cmd.CombinedOutput()
```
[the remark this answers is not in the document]
brianmcgee: My preference is for the `cmd.Output()` approach and type casting the `ExitError` to extract the stderr if needed. More straightforward and doesn't involve a second thread of execution.

config/config.go
brianmcgee: Multiline output will force the resolution to an absolute path to bork later on. That being said, it wouldn't hurt to catch this specifically and provide pointed feedback that the script or command being used is not behaving as expected.

config/config.go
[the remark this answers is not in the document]
brianmcgee: > I agree that's simpler, although both approaches should only need one thread; go-routines optimise for concurrency without creating unnecessary threads IIUC.

I wasn't speaking in terms of actual threads, more about keeping the execution path linear. 

You make a fair point though about making stderr available for diagnostic info regardless of exit code. The only way to achieve that is like you shared above.

config/config.go
```diff
@@ -263,6 +258,105 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+	// split the command first, resolving any '' and "" entries
+	parts, splitErr := shlex.Split(cfg.TreeRootCmd)
+	if splitErr != nil {
+		return "", fmt.Errorf("failed to parse tree-root-cmd: %w", splitErr)
+	}
+
+	// set a reasonable timeout of 2 seconds to wait for the command to return
+	// it shouldn't take anywhere near this amount of time unless there's a problem
+	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+	defer cancel()
```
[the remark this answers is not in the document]
brianmcgee: Updated to explain what happened in the error.

config/config.go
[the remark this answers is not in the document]
brianmcgee: Made some changes and added some tests

config/config.go
brianmcgee: Added a test for this and an explicit error

cmd/root_test.go
jfly: Urg, sorry, I was only complaining about the nested quotes. (I moved this comment around after I wrote it.)

config/config.go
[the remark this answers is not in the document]
jfly: I haven't read the updated code yet, but the simplest way to make sure the user sees stderr is to just not capture it.

config/config.go
```diff
@@ -263,6 +262,167 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+	if cfg.TreeRootCmd != "" {
+		count++
+	}
+
+	if cfg.TreeRootFile != "" {
+		count++
+	}
+
+	if count > 1 {
+		return errors.New("only one of tree-root, tree-root-cmd or tree-root-file can be specified")
```
jfly, line 286: nit:

```suggestion
		return errors.New("at most one of tree-root, tree-root-cmd or tree-root-file can be specified")
```

config/config.go
```diff
@@ -263,6 +262,167 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+		// otherwise, some other kind of error occurred
+		return "", fmt.Errorf("failed to execute tree-root-cmd: %w", cmdErr)
+	}
+
+	// trim the output and check it's not empty
+	treeRoot := strings.TrimSpace(string(output))
+
+	if treeRoot == "" {
+		return "", fmt.Errorf("empty output received after executing tree-root-cmd: %s", cfg.TreeRootCmd)
+	}
```
jfly, line 417: I personally would prefer to put this check up a layer in `determineTreeRoot`. If somebody explicitly specifies `--tree-root ""`, I'd also like them to get this error, rather than having it fall back to the current directory.

jfly (approved):
Nice work!

config/config.go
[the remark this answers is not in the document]
MattSturgeon: > but the simplest way to make sure the user sees stderr is to just not capture it.

You can do `cmd.Stderr = os.Stderr` to avoid capturing it. Definitely much simpler than dealing with pipes and concurrency.

The only (potential) disadvantage is that you can't "wrap" the command's stderr with treefmt's logger.

Probably not a big deal, but it would mean users may not understand that the uncaptured output is coming from `treefmt: tree-root-cmd:`, since it'd print without any context prefix. It'd also print unconditionally, regardless of log level.

I don't think either approach is "better," just a design decision to be made.

config/config.go
MattSturgeon: > If somebody explicitly specifies `--tree-root " "`

I was about to comment that `""` would be treated as `tree-root` being undefined, however if a whitespace-only value is defined then you're right, it _would_ run into this edge-case.

config/config_test.go
```diff
@@ -457,6 +488,48 @@ func TestTreeRootFile(t *testing.T) {
...
+		})
+	}
+
+	tempDir := t.TempDir()
+	as.NoError(os.MkdirAll(filepath.Join(tempDir, "foo"), 0o755))
+	as.NoError(os.MkdirAll(filepath.Join(tempDir, "bar"), 0o755))
+
+	// default with no flag, env or config
+	// should match the absolute path of the directory in which the config file is located
+	checkValue(filepath.Dir(v.ConfigFileUsed()))
```
MattSturgeon, line 509: Should there also be a variant of this test-case for when `tempDir` is a git repo?

In that scenario we'd expect `checkValue(tempDir)`, right?

config/config.go
```diff
@@ -263,6 +262,167 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+	// set git-based tree root command if the walker is git and no tree root has been specified
+	if cfg.Walk == walk.Git.String() && count == 0 {
+		cfg.TreeRootCmd = "git rev-parse --show-toplevel"
+
+		logger.Infof(
+			"git walker enabled and tree root has not been specified: defaulting tree-root-cmd to '%s'",
+			cfg.TreeRootCmd,
+		)
+	}
```
MattSturgeon, line 297: This new behaviour should probably be documented too.

I believe the docs currently state `tree-root`:

> Defaults to the directory containing the config file.

Now that should say something like:

> Defaults to the root of the current git worktree. If not in a git repo, defaults the directory containing the config file.

config/config.go
jfly: Oops, good point. I didn't remember we're treating "unset" and "set to empty string" as the same thing here.

I still do feel like it would make more sense to do this check as part of the "is this potential root a valid path? let's normalize it" codepath, and not in the implementation of a specific root discovery mechanism.

MattSturgeon (commented):
LGTM overall. Thanks again for working on this!

A few minor things outstanding most of which you probably don't need to block on if you don't want to:

- [ ] Docs for how tree-root-cmd is parsed (https://github.com/numtide/treefmt/pull/571#discussion_r2053825300)
- [ ] Maybe docs for _where_ tree-root-cmd is run (https://github.com/numtide/treefmt/pull/571#discussion_r2053847104)
- [ ] Update docs for `tree-root`'s default (https://github.com/numtide/treefmt/pull/571#discussion_r2054889429)
- [ ] Maybe also checking the _trimmed_ treeRoot is non-empty for explicit `tree-root` defs (https://github.com/numtide/treefmt/pull/571#discussion_r2054742625)
- [ ] Packaging nits
        (https://github.com/numtide/treefmt/pull/571#discussion_r2054848662)
        (https://github.com/numtide/treefmt/pull/571#discussion_r2054849390)
- [ ] A couple more nits below:

config/config.go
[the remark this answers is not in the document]
brianmcgee: On second thought I feel this is self evident, and it's just adding to an already verbose description for `tree-root-cmd` in the cli.

config/config.go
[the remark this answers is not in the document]
brianmcgee: As @MattSturgeon pointed out, capturing it allows us to prefix log it and make it clear where it came from. We already do similar when executing the formatters.

nix/packages/treefmt/default.nix
```diff
@@ -117,7 +117,7 @@ in
           env.CGO_ENABLED = 1;
           buildPhase = ''
             HOME=$TMPDIR
-            go test -race -covermode=atomic -coverprofile=coverage.out -v ./...
+            go test -race -covermode=atomic -coverprofile="coverage.out" -v ./...
```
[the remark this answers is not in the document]
brianmcgee: I'd like to address the packaging nits in a separate PR.

config/config.go
brianmcgee: I've updated the description.

config/config.go
```diff
@@ -263,6 +258,105 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+	log.Debugf("tree root: %s", cfg.TreeRoot)
+
+	return nil
+}
+
+func execTreeRootCmd(cfg *Config) (string, error) {
+	// split the command first, resolving any '' and "" entries
+	parts, splitErr := shlex.Split(cfg.TreeRootCmd)
```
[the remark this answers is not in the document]
brianmcgee: I've updated the description and the comments in the default init toml file.

config/config.go
brianmcgee: Updated

config/config.go
brianmcgee: I'm not quite following what is being advocated for here.

config/config.go
```diff
@@ -263,6 +262,167 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+	// trim the output and check it's not empty
+	treeRoot := strings.TrimSpace(string(output))
+
+	if treeRoot == "" {
+		return "", fmt.Errorf("empty output received after executing tree-root-cmd: %s", cfg.TreeRootCmd)
+	}
+
+	if strings.Contains(treeRoot, "\n") {
+		return "", fmt.Errorf("tree-root-cmd cannot output multiple lines: %s", cfg.TreeRootCmd)
```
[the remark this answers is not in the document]
brianmcgee: I've changed it to log stdout as an error:

```console
    root_test.go:1629: treefmt --tree-root-cmd bash -c 'echo /tmp/TestTreeRootCmd2421694470/001 && echo /tmp/TestTreeRootCmd2421694470/001'
    root_test.go:1629: 
        ERRO tree-root-cmd | stdout: 
        /tmp/TestTreeRootCmd2421694470/001
        /tmp/TestTreeRootCmd2421694470/001
        
        Error: failed to load config: failed to determine tree root: tree-root-cmd cannot output multiple lines: bash -c 'echo /tmp/TestTreeRootCmd2421694470/001 && echo /tmp/TestTreeRootCmd2421694470/001'
      
```

config/config.go
MattSturgeon: > IMO this begs the question of whether trimming stdout is the correct approach anyway.

In hindsight, I believe the most correct approach here would be to split on `\n`, then remove the empty lines.

You can then assert that the remaining line count is 1, and also use the lines slice to log the stdout line-by-line.

```go
lines := strings.Split(stdout, "\n")
nonEmptyLines := slices.DeleteFunc(lines, func(line string) bool {
	return line == ""
}

switch (len(nonEmptyLines)) {
case 1:
	// no-op

case 0:
	return "", fmt.Errorf("empty output received after executing tree-root-cmd: %s", cfg.TreeRootCmd)

default:
	// Log all stdout line-by-line, including the blank lines
	logger := log.WithPrefix("tree-root-cmd | stdout")
	for _, line := range lines {
		logger.Error(line)
	}
	return "", fmt.Errorf("tree-root-cmd cannot output multiple lines: %s", cfg.TreeRootCmd)
}
```

---

As said before, trimming is fine for the initial impl, I can open a follow up PR for this if preferred.

brianmcgee: All mechanisms must pass through the same stage of realising their output as absolute paths, which should throw an error for any malformed paths. Trimming the output of the command execution and enforcing a single line seems prudent, specifically for the `tree-root-cmd` pathway.

config/config_test.go
MattSturgeon: I guess its fine to omit that test if this is intended to be more of a unit test, and there's integration tests elsewhere that cover the _"find tree root without any `tree-root-*` setting defined, while in a git repo"_ scenario?

I'm not too familiar with your test suite, but it'd be nice if that scenario had _some_ coverage somewhere :slightly_smiling_face:

MattSturgeon (commented):
Sorry to review again with such minor comments... Despite the noise, the PR _is_ looking really good!

config/config.go
```diff
@@ -101,12 +110,19 @@ func SetFlags(fs *pflag.FlagSet) {
 	)
 	fs.String(
 		"tree-root", "",
-		"The root directory from which treefmt will start walking the filesystem (defaults to the directory "+
-			"containing the config file). (env $TREEFMT_TREE_ROOT)",
+		"The root directory from which treefmt will start walking the filesystem. "+
+			"Defaults to the root of the current git worktree. If not in a git repo, defaults to the directory "+
+			"containing the config file. (env $TREEFMT_TREE_ROOT)",
```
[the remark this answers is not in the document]
brianmcgee: Updated

brianmcgee:
> Sorry to review again with such minor comments... Despite the noise, the PR _is_ looking really good!

I don't mind, I'd rather get it right.

MattSturgeon (approved):
Only things outstanding for me are

- [ ] Removing empty lines instead of arbitrarily trimming stdout
   (https://github.com/numtide/treefmt/pull/571#discussion_r2056030549)
- [ ] Having a test case for git-repo & no defined `tree-root` option
   (https://github.com/numtide/treefmt/pull/571#discussion_r2054880118)

Neither of these are _blocking_ IMO, so I approve :tada: 

(I haven't tested locally, but the diff & tests all look good)

config/config.go
MattSturgeon: > This is the part that bothers me: IIRC, we're using go's built-in `Path.abs` to do this, which accepts empty strings rather than erroring.

I believe it is impossible to define a _truly_ empty tree-root path; attempting to set `tree-root = ""` will be treated as undefined. Returning an empty string from `tree-root-cmd` will be caught here.

As for blank/whitespace paths, a whitepace string is technically a valid relative path. On some filesystems you can name a file/dir `" "` (if you really wanted to).

This is one reason why the current impl splits stdout into lines, instead of just trimming it.

PR #572, opened 2025-04-11
fix: resolve symlinks in tree root

jacobkoziej:
The following resolves symlinks in the tree root path to avoid false-positives where paths (with symlinks) specified inside the tree root are marked outside.


jacobkoziej:
Looks like `TestTreeRoot` fails with this change (which is to be expected). I think the reasonable course of action is to re-write the test so that temporary directories that exist on disk get used. I'd offer to do this, but I'm not familiar enough with golang or the testing framework treefmt uses :(

jfly (commented):
Thanks, @jacobkoziej!

> I think the reasonable course of action is to re-write the test so that temporary directories that exist on disk get used. 

That sounds reasonable to me. See this test for an example of how to create a temp dir for a test: https://github.com/numtide/treefmt/blob/9399cd69f9860bf62fd42479591e0ae39a1b7636/config/config_test.go#L422.

We should also make sure the user experience is OK when a non-existent directory is specified.

PR #586, opened 2025-05-13
Add a Stdin Specification

jfly:
This described the Stdin Specification, and adds a new `formatter.stdin_options` config to enable treefmt to invoke formatters in "stdin mode".

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -2257,6 +2257,31 @@ help:
 `, string(out))
 		}),
 	)
+
+	// Format with embed-path. `embed-path` prefixes the file with its path. If
+	// this formatter didn't support the Stdin Specification, then we'd see a
+	// temp path get added instead of the true path to the file.
+	contents = `
+The formatter should add the path above this line.
+`
+	os.Stdin = test.TempFile(t, "", "stdin", &contents)
+
+	treefmt(t,
+		withArgs("-vvv", "--stdin", "path/to/foo.embed-path"),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 1,
+			stats.Matched:   1,
+			stats.Formatted: 1,
+			stats.Changed:   1,
+		}),
+		withStdout(func(out []byte) {
+			as.Equal(`# path/to/foo.embed-path
+
+The formatter should add the path above this line.
+`, string(out))
+		}),
+	)
 }
 
 func TestDeterministicOrderingInPipeline(t *testing.T) {
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -2288,12 +2288,44 @@ help:
 		}),
 	)
 
+	// Format with embed-path. `embed-path` prefixes the file with its path. If
+	// this formatter didn't support the Stdin Specification, then we'd see a
+	// temp path get added instead of the true path to the file.
+	contents = `
+The formatter should add the path above this line.
+`
+	os.Stdin = test.TempFile(t, "", "stdin", &contents)
+
+	treefmt(t,
+		withArgs("-vvv", "--stdin", "path/to/foo.embed-path"),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 1,
+			stats.Matched:   1,
+			stats.Formatted: 1,
+			stats.Changed:   1,
+		}),
+		withStdout(func(out []byte) {
+			as.Equal(`# path/to/foo.embed-path
+
+The formatter should add the path above this line.
+`, string(out))
+		}),
+	)
+
 	// Try from a subdirectory, using .. to get to a parent directory that is
 	// still inside the project root.
 	//
 	// Note: we called `test.ChangeWorkDir` at the start of the test, which
 	// will restore the working directory during test cleanup.
 	t.Chdir("go")
+
+	contents = `
+# print this message
+help:
+        just --list --list-submodules --unsorted
+
+`
 	os.Stdin = test.TempFile(t, "", "stdin", &contents)
 	treefmt(t,
 		withArgs("--stdin", "../foo/justfile"),
@@ -2821,6 +2853,14 @@ func treefmt(
 		os.Stdout = stdout
 		os.Stderr = stderr
 		log.SetOutput(stderr)
+	}()
+
+	// treefmt may change the configured log level when we invoke it. Be sure
+	// to restore the previous level.
+	ogLogLevel := log.GetLevel()
+
+	defer func() {
+		log.SetLevel(ogLogLevel)
 	}()
 
 	// run the command
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,12 @@ type Formatter struct {
 	Command string `mapstructure:"command" toml:"command"`
 	// Options are an optional list of args to be passed to Command.
 	Options []string `mapstructure:"options,omitempty" toml:"options,omitempty"`
+	// StdinOptions are an optional list of args used to invoke the formatter
+	// in stdin mode (where it reads the buffer to format from stdin rather
+	// than a file). Any occurrences of `$path` will be replaced with the "advisory path"
+	// of the "file" being formatted (useful for formatters whose behavior
+	// depend on the path of the file being formatted).
+	StdinOptions []string `mapstructure:"stdin-options,omitempty" toml:"stdin-options,omitempty"`
 	// Includes is a list of glob patterns used to determine whether this Formatter should be applied against a path.
 	Includes []string `mapstructure:"includes,omitempty" toml:"includes,omitempty"`
 	// Excludes is an optional list of glob patterns used to exclude certain files from this Formatter.
--- a/config/config.go
+++ b/config/config.go
@@ -57,7 +57,7 @@ type Formatter struct {
 	Command string `mapstructure:"command" toml:"command"`
 	// Options are an optional list of args to be passed to Command.
 	Options []string `mapstructure:"options,omitempty" toml:"options,omitempty"`
-	// StringOptions are an optional list of args used to invoke the formatter
+	// StdinOptions are an optional list of args used to invoke the formatter
 	// in stdin mode (where it reads the buffer to format from stdin rather
 	// than a file). Any occurrences of `$path` will be replaced with the "advisory path"
 	// of the "file" being formatted (useful for formatters whose behavior
--- a/config/config.go
+++ b/config/config.go
@@ -57,6 +57,12 @@ type Formatter struct {
 	Command string `mapstructure:"command" toml:"command"`
 	// Options are an optional list of args to be passed to Command.
 	Options []string `mapstructure:"options,omitempty" toml:"options,omitempty"`
+	// StringOptions are an optional list of args used to invoke the formatter
+	// in stdin mode (where it reads the buffer to format from stdin rather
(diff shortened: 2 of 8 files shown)


docs/site/reference/formatter-spec.md
```diff
@@ -60,3 +60,54 @@ outputs.
...
+!!! note
+
+    This CLI _MUST_ be implemented in a way that is compatibile with the
+    vanilla [Formatter Specification](#1-files-passed-as-arguments).
+
+#### 2. Print to stdout, do not assume file is present on filesystem
+
+When in stdin mode, the formatter:
+
+1. **MUST** print the formatted file to stdout.
```
jfly, line 110: This is a lie: it's OK for the formatter to fail to format, but it probably should (must?):

- Print an explanation *why* to stderr
- Exit non-zero

This made me realize the current formatter specification doesn't have anything to say about subprocess exit codes, but treefmt does [check the exit code of the formatter it invokes](https://github.com/numtide/treefmt/blob/v2.2.1/format/formatter.go#L107). We should probably mention this in the spec?

docs/site/reference/formatter-spec.md
```diff
@@ -60,3 +60,54 @@ outputs.
 ### 4. Reliable
 
 We expect the formatter to be reliable and not break the semantics of the formatted files.
+
+## Stdin Specification
```
jfly, line 64: Should I actually promote this to a top level file? That is, `docs/site/reference/stdin-spec.md`?

docs/site/reference/formatter-spec.md
```diff
@@ -60,3 +60,54 @@ outputs.
...
+  stdin mode, the formatter reads file contents from stdin rather than the
+  filesystem.
+- `[...<files>]` is one or more files given to the formatter for processing. If
+  `--stdin` is specified, then exactly 1 must be present, and is treated as a
+  filename.
+
+Example:
+
+```
+$ echo "{}" | nixfmt --stdin path/to/file.nix
```
jfly, line 98: This is a bad example, as [nixfmt doesn't actually behave this way yet](https://github.com/NixOS/nixfmt/issues/305).

Do folks know if there's an existing formatter (besides treefmt itself) that implements this spec?

jfly:
@brianmcgee @zimbatm, is there anything I can do to help move this forward?

brianmcgee:
It's unclear to me if this documents the current behaviour or the desired behaviour. If it's the desired behaviour, isn't there some work needing to be done in `treefmt` to support that before we publish a spec?

jfly:
@brianmcgee, friendly ping

brianmcgee (approved):
I agree with some of your own musings in the comments you'd left. Otherwise, it seems like a sane spec and one which should be easy enough for formatters to follow. 

I also like that this is opt-in.

brianmcgee:
P.S. Thanks for taking the time to flesh this out, and sorry for taking a while to get around to reviewing this.

docs/site/reference/formatter-spec.md
brianmcgee: Hmmm, then I can't say I know of any formatter that actually conforms to this yet. Sounds like `ormolu` might be close?

docs/site/reference/formatter-spec.md
jfly: I've updated the formatter spec in https://github.com/numtide/treefmt/pull/610.

I've updated the stdin spec with the exact same rule as well.

docs/site/reference/formatter-spec.md
jfly: Bummer. I'm going to leave this as is, then. Hopefully we can change nixfmt to honor this spec before we merge this.

jfly: Discussed at today's Nix Formatting team meeting: we like `--stdin-filepath=<path>` for 2 reasons:

1. This is path, not a filename.
2. It's more obvious that this expects exactly 1 path, rather than `[files...]`

docs/site/reference/stdin-spec.md
```diff
@@ -0,0 +1,63 @@
...
+3. **MUST NOT** write to the given path on the filesytem. It _MAY_ write to
+   temporary files elsewhere on disk, but _SHOULD_ clean them up when done.
+
+### 3. Exit nonzero on error
+
+When something goes wrong when formatting (for example, the source code is
+syntactically invalid), the formatter **MUST** exit nonzero.
+
+The formatter _SHOULD_ print useful information about what went wrong to
+stderr.
```
jfly, line 63: @infinisil points out that this is redundant with the vanilla formatter spec. Remove.

docs/site/reference/formatter-spec.md
jfly: I've updated the spec to use `--stdin-filepath`

infinisil (approved):
LGTM!

docs/site/reference/stdin-spec.md
```diff
@@ -0,0 +1,51 @@
...
+outline: deep
+---
+
+# Stdin Specification
+
+Formatters **MAY** also implement the Stdin Specification, which allows
+formatting "virtual files" passed via stdin.
+
+A formatter **MUST** implement the Stdin Specification if its formatting behavior
+can depend on the name of the file being formatted.
```
zimbatm, line 11: this is a bit hard to read

docs/site/reference/stdin-spec.md
```diff
@@ -0,0 +1,51 @@
...
+
+Where:
+
+- `<command>` is the name of the formatting tool.
+- `[options]` are any number of flags and options that the formatter accepts.
+- `--stdin-filepath <path>` is an optional flag that puts the formatter in
+  "stdin mode". In stdin mode, the formatter reads file contents from stdin
+  rather than the filesystem.
+    - The formatter _MAY_ understand `--stdin-filepath=<path>` as well, but **MUST**
+      understand the space separated variant.
```
zimbatm, line 35: It would make sense to outline the purpose of the `<path>` argument explicitly. Both you and I know why it's there, but maybe not the reader.

zimbatm: If the path argument is not needed by the tool, it's also valid to ignore it.

docs/site/reference/stdin-spec.md
```diff
@@ -0,0 +1,51 @@
...
+A formatter **MUST** implement the Stdin Specification if its formatting behavior
+can depend on the name of the file being formatted.
+
+## Rules
+
+In order for the formatter to comply with this spec, it **MUST** implement the
+vanilla [Formatter Specification](/reference/formatter-spec), and additionally
+satisfy the following:
+
+### 1. `--stdin-filepath` flag
```
zimbatm, line 19:
```suggestion
### 1. `--stdin` flag
```
KISS?

zimbatm (commented):
Some minor things, but directionally it makes sense.

docs/site/reference/stdin-spec.md
```diff
@@ -0,0 +1,51 @@
+---
+outline: deep
+---
+
+# Stdin Specification
+
+Formatters **MAY** also implement the Stdin Specification, which allows
+formatting "virtual files" passed via stdin.
```
zimbatm, line 8: How is this support expressed in the treefmt.toml config?

docs/site/reference/stdin-spec.md
jfly: I've taken a stab at this. Please take a look.

docs/site/reference/stdin-spec.md
```diff
@@ -0,0 +1,55 @@
...
+- `[options]` are any number of flags and options that the formatter accepts.
+- `--stdin-filepath <path>` is an optional flag that puts the formatter in
+  "stdin mode". In stdin mode, the formatter reads file contents from stdin
+  rather than the filesystem.
+    - The formatter _MAY_ alter its behavior based on the given `<path>`. For
+      example, if a there are different formatting rules in different
+      directories. If the formatter's behavior doesn't depend on the given
+      `<path>`, it's ok to ignore it.
+    - The formatter _MAY_ understand `--stdin-filepath=<path>` as well, but **MUST**
+      understand the space separated variant.
```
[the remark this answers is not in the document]
jfly: > According to that documentation, any go-based formatter cannot support this without swapping away from the standard library's flags package

Sorry, I don't follow this point. From [the page](https://pkg.go.dev/flag#hdr-Command_line_flag_syntax) you linked to, I see this:

> The following forms are permitted:
> 
> ```
> --flag   // double dashes are also permitted
> ```

What am I missing?

> Basically please don't make me do that. [...] it's worth being flexible so this can be useful without waiting 2 years for a bunch of CLI flag adjustments to go through.

I hear you. We will definitely need a compatibility story, but it's TBD if that lives in treefmt itself, or treefmt-nix.

> If the tool does support stdin mode, I don't see why you would ever want to run it in non-stdin mode. [...] Edit: Unless you're running multiple files per formatter instance, in which case rustfmt does not duplicate work and will keep its code in cache better

This is exactly why: invoking the formatter N / B times (where B is our batch size) rather than N times.

> you still may not want this if some other formatter is single threaded

This is a good question. We probably kind of assume that formatters are single threaded.

config/config.go
```diff
@@ -56,6 +56,12 @@ type Formatter struct {
 	Command string `mapstructure:"command" toml:"command"`
 	// Options are an optional list of args to be passed to Command.
 	Options []string `mapstructure:"options,omitempty" toml:"options,omitempty"`
+	// StringOptions are an optional list of args used to invoke the formatter
```
manuelteillet: StringOptions -> StdinOptions

jfly: Thank you, fixed.

jbms:
This addresses a pretty significant limitation in treefmt --- is there something holding up this change?  It looks pretty reasonable to me.

jfly:
I'm happy with the state of this, just need approval from @brianmcgee.

brianmcgee (approved):

brianmcgee:
Needs a rebase, but I've approved the PR as I think this is a sane approach. Thanks for driving this, I've not had much time for treefmt of late.

cmd/root_test.go
```diff
@@ -2288,12 +2288,44 @@ help:
...
+	)
+
 	// Try from a subdirectory, using .. to get to a parent directory that is
 	// still inside the project root.
 	//
 	// Note: we called `test.ChangeWorkDir` at the start of the test, which
 	// will restore the working directory during test cleanup.
 	t.Chdir("go")
+
+	contents = `
```
jfly, line 2323: Seems unrelated, but this test relied upon the value of `contents` from a previous test (which was shadowed by the test I added). Rather than carefully shuffle things around to keep them happy, I opted to specify `contents` explictly. IMO, this results in a better test.

cmd/root_test.go
```diff
@@ -2823,6 +2855,14 @@ func treefmt(
 		log.SetOutput(stderr)
 	}()
 
+	// treefmt may change the configured log level when we invoke it. Be sure
```
jfly, line 2858: Discovered this when I added my test which calls `treefmt -vvv`, which affected the subsequent test.

jfly:
Thanks @brianmcgee! I've rebased, and did another review over the code. I opted to clarify the terminology we've invented here (specifically "virtual file" and "advisory path"), and I've updated the code and docs to use the terms consistently).

PTAL!

PR #588, opened 2025-05-19
Fix defaulting of `tree-root-cmd`.

brianmcgee:
Refines the logic for `tree-root-cmd`, handling the case where `walk` is `auto` and adds some missing docs.

Relates to https://github.com/NixOS/nixpkgs/pull/407681

CC @MattSturgeon


config/config.go
```diff
@@ -290,14 +291,35 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
...
+
+			// if it is, set the default git tree root command
+			if isGitRepo {
+				setGitRootCmd()
+			}
+
+		case cfg.Walk == walk.Git.String():
+			setGitRootCmd()
+
+		default: // otherwise, do nothing
```
MattSturgeon, line 321: It didn't before, but should this log anything?

E.g. "No 'tree-root', 'tree-root-file', or 'tree-root-cmd' specified and not in a git repo. Assuming working directory is the tree root: %s"

config/config_test.go
```diff
@@ -617,12 +617,14 @@ func TestWorkingDirectory(t *testing.T) {
...
 	// env override
-	t.Setenv("TREEFMT_WORKING_DIR", "/fizz/buzz/..")
-	checkValue("/fizz")
+	cwd = t.TempDir()
+	t.Setenv("TREEFMT_WORKING_DIR", cwd+"/buzz/..")
+	checkValue(cwd)
 
 	// flag override
-	as.NoError(flags.Set("working-dir", "/flip/flop"))
-	checkValue("/flip/flop")
```
MattSturgeon, line 625: I don't understand why this test was changed?

Is it made redundant by the new test?

git/git.go
MattSturgeon, line 1: CI is complaining that this is untested. Is it worth having functional tests for "is a git worktree", to check it handles things like:

1. Being in a bare repo with no worktree
2. Being in a "normal" repo with a `.git` dir
3. Being in a worktree with a `.git` file
4. Being entirely outside of a git repo

?

I assume some of these scenarios are why you have code checking _both_ `exitErr` _and_ not having `"true"` in stdout.

config/config.go
```diff
@@ -290,14 +291,35 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
...
+			// check if the directory containing the config file is a git repo
+			isGitRepo, err := git.IsInsideWorktree(cfg.WorkingDirectory)
+			if err != nil {
+				return fmt.Errorf("failed to check if working directory is a git repo: %w", err)
+			}
+
+			// if it is, set the default git tree root command
+			if isGitRepo {
+				setGitRootCmd()
+			}
```
MattSturgeon, line 316: Would it be simpler, instead of a) checking if we're in a git directory and b) setting a default `tree-root-cmd` if we are, to instead do something like:

If no `tree-root`, `tree-root-file`, or `tree-root-cmd` are set; then a) attempt to set `tree-root` directly by running `git rev-parse --show-toplevel` b) silently fail if this fails.

Without explicitly checking that we're actually in a git repo via `git rev-parse --is-inside-work-tree`.

It seems to me, the logic of checking "are we in a git repo" is the same logic we would need to attempt to run `git rev-parse --show-toplevel` without knowing (or caring) if we're actually in a git repo.

config/config_test.go
brianmcgee: With the new logic I needed these to be real directories.

config/config.go
brianmcgee: It's a fair simplification. I'll have a look in a bit.

brianmcgee: Much simpler now. I'll look at beefing up test coverage this evening.

config/config.go
```diff
@@ -315,14 +306,26 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
...
 
 	default:
 		// no tree root was specified
+		logger.Debugf("no tree root specified")
+
+		// attempt to resolve the tree root with git
+		if cfg.Walk == walk.Auto.String() || cfg.Walk == walk.Git.String() {
+			if cfg.TreeRoot, err = execTreeRootCmd(git.TreeRootCmd, cfg.WorkingDirectory); err != nil {
+				// fail silently
+				logger.Debugf("failed to resolve tree root with git: %v", err)
```
MattSturgeon, line 321: Should this be debug or info? My instinct is that "attempted and failed to find a git tree root" is more of an info-level log, but no strong feelings.

Should it fail silently for `walk == "auto"` but return an error for `walk == "git"`? Not too important from my perspective, and I'm happy with either answer; just raising the question!

Since `execTreeRootCmd` has its own internal error handling, maybe it could have a new function parameter (`required`, `optionsl`, etc)? That way:
- `execTreeRootCmd` could decide internally whether a failure is bad enough to fail, even when `required=false` or `optional=true`
- this calling code could supply the `required` arg as `cfg.Walk == walk.Git.String()` _if_ you decide that case shouldn't fail silently. (Or `optional` as `cfg.Walk == walk.Auto.String()`).

Only if you feel this additional complexity is worth it, or simplifies things elsewhere.

config/config.go
```diff
@@ -315,14 +306,26 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
...
+			} else {
+				logger.Debugf("resolved tree root with git: %s", cfg.TreeRoot)
```
MattSturgeon, line 323: Again, maybe should be an info-level log? Up to you.

MattSturgeon (commented):
Looks good IMO, thanks for working on this!

One suggestion for another minor refactor, if desired:

config/config.go
brianmcgee: Made some improvements on the logging.

git/git.go
brianmcgee: I'll follow this up in a separate PR. I want to get this fix out before it starts generating unexpected behaviour for users.

PR #591, opened 2025-05-22
refactor: simplify one line of nix code

MattSturgeon:
Sorry, this was bugging me last time I looked at this code.

We don't actually care if `__functionArgs.treefmt` is null (it'll always be `true` or `false`). We actually care if the attr is present.

IIRC, when I first wrote that line I didn't know you could use a nested attrpath on the RHS of the `?` operator, so I resorted to misusing `or` instead.


brianmcgee (approved):

zimbatm:
TIL

PR #593, opened 2025-05-23
fix: force-remove db file when clearing cache

### LINKED ISSUE #592: `failed to open cache db: timeout`
### Describe the bug

It is possible to get the `treefmt` cache database into a bad state where all `treefmt` operations fail, including `treefmt --clear-cache`!

### To Reproduce

We do not have a reproducer available, but we will update this ticket if we find one.

This issue manifests with errors instead of formatting:

```
$ treefmt
Error: failed to open cache: failed to open cache db: timeout
$ treefmt --clear-cache
Error: failed to open cache: failed to open cache db: timeout
```

### Expected behavior

When `treefmt` is run, it should work instead of crashing.

If the cache database is corrupted, it should be deleted instead of crashing `treefmt`.

`treefmt --clear-cache` should never fail because the cache database is unavailable; the database should be deleted, not opened.

### System information

macOS, treefmt v2.2.1

### Additional context

The issue was resolved by manually deleting `~/Library/Caches/treefmt`. This was challenging because the cache directory is not documented, and the source code implies it lives in `$XDG_CACHE_HOME`, which is not how `treefmt` behaves on macOS.

brianmcgee:
It also improves logging of cache file location when the db cannot be opened.

Fixes #592


cmd/format/format.go
```diff
@@ -77,6 +77,13 @@ func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string)
 		}()
 	}
 
+	// clear the cache if desired before attempting to open its\
+	if cfg.ClearCache && !cfg.NoCache { // clear the cache if desired
```
jfly, line 81: nit: comments got a bit wonky here:

```suggestion
	// clear the cache if desired before attempting to open it
	if cfg.ClearCache && !cfg.NoCache {
```

Also, IMO it's weird to not clear the cache if `--no-cache` is specified. I'd suggest one of the following instead:

1. Just clear the cache, OR
2. Print out a warning explaining why we're not clearing the cache despite `--clear-cache` being specified.

cmd/root_test.go
```diff
@@ -2141,6 +2142,79 @@ func TestProjectRootIsSymlink(t *testing.T) {
...
+		treefmt(t,
+			withArgs("--formatters", "slow"),
+			withConfig(configPath, cfg),
+			withNoError(t),
+		)
+
+		return nil
+	})
+
+	time.Sleep(time.Second)
```
jfly, line 2183: Why the sleep? Is it to give the previous treefmt a moment to start up (and lock the boldb cache) before we start the next treefmt?

Am I understanding this correctly?

- At time 0: We start a treefmt ("slow") that will grab the bolddb cache, and sleep 2 seconds before formatting and releasing the bolddb cache.
- Sleep 1 second
- At time 1: We start a second treefmt ("echo") that will wait up to 1 second to grab the boltdb cache.
- At time 2: "slow" is finishing up while "echo" is just about ready to give up on waiting for the cache. 

IIUC, this feels racy to me, it's not a sure thing that "echo" will always hit the "failed to open cache" error. I'd fix this by just making "slow" quite a bit slower (5 or 10 seconds feels like plenty to me).

cmd/root_test.go
```diff
@@ -2141,6 +2142,79 @@ func TestProjectRootIsSymlink(t *testing.T) {
...
+
+	treefmt(t,
+		withArgs("-c", "--formatters", "echo"),
```
jfly, line 2210: nit, feel free to ignore:

```suggestion
		withArgs("--clear-cache", "--formatters", "echo"),
```

walk/cache/cache.go
```diff
@@ -56,28 +64,16 @@ func PathsBucket(tx *bolt.Tx) *bolt.Bucket {
...
-		err := deleteAll(PathsBucket(tx))
-		if err != nil {
-			return fmt.Errorf("failed to clear cache: %w", err)
-		}
-
-		return nil
-	})
-	if err != nil {
-		return fmt.Errorf("failed to clear cache: %w", err)
+	// force-remove the db file
```
jfly, line 74: nit: What does "force" mean here?

walk/cache/cache.go
```diff
@@ -14,26 +15,33 @@ const (
...
+	path, err := Path(root)
+	if err != nil {
+		return nil, err
 	}
 
 	// open db
 	db, err := bolt.Open(path, 0o600, &bolt.Options{Timeout: 1 * time.Second})
 	if err != nil {
-		return nil, fmt.Errorf("failed to open cache db: %w", err)
+		return nil, fmt.Errorf("failed to open cache db at %s: %w", path, err)
```
jfly, line 44: (Not a blocker for this PR) Does bbolt have any mechanism for reporting information on *what* has the cache locked?

cmd/root_test.go
```diff
@@ -2141,6 +2142,79 @@ func TestProjectRootIsSymlink(t *testing.T) {
...
+		withConfig(configPath, cfg),
+		withNoError(t),
+	)
+
+	as.NoError(eg.Wait())
```
jfly, line 2215: IIUC, at this point, the "slow" formatter is still running (presumably it is operating on an open file inode that no longer has a name). Is it possible it will re-create the cache when it exits?

walk/cache/cache.go
```diff
@@ -56,28 +64,16 @@ func PathsBucket(tx *bolt.Tx) *bolt.Bucket {
...
-		if err != nil {
-			return fmt.Errorf("failed to clear cache: %w", err)
-		}
-
-		return nil
-	})
-	if err != nil {
-		return fmt.Errorf("failed to clear cache: %w", err)
+	// force-remove the db file
+	if err = os.Remove(path); !(err == nil || os.IsNotExist(err)) {
```
jfly, line 75: Do we support Windows? Windows doesn't let you delete files if they're still open in another process, right?

jfly (approved):
Thanks! I agree that just nuking the db feels like the right move, rather than adding yet another knob for "yes just do it".

cmd/root_test.go
brianmcgee: I thought I had reduced that to 500ms. Made the change.

walk/cache/cache.go
brianmcgee: Updated the comment to make things clearer.

walk/cache/cache.go
brianmcgee: Don't know, worth looking into.

Issue #594, opened 2025-05-25
`treefmt .` at the root of the project fails if the project root is a symlink


jfly:
**Describe the bug**

If the root of your project is a symlink, and you try running `treefmt .` at the root of the repo, you'll get a segfault.

**To Reproduce**

1. Create a project `./prj` with a working treefmt configuration.
2. Create a symlink `./prj-link`: `ln -s prj prj-link`.
3. `cd prj-link`
4. `treefmt .`
   ```shell
   $ treefmt .
   panic: runtime error: invalid memory address or nil pointer dereference
   [signal SIGSEGV: segmentation violation code=0x1 addr=0x20 pc=0x6c0548]

   goroutine 1 [running]:
   github.com/numtide/treefmt/v2/walk.(*CompositeReader).Read(0xc000214180, {0x9775a8?, 0xc0002042a0?}, {0xc00022c008?, 0x1?, 0x85bf20?})
       github.com/numtide/treefmt/v2/walk/walk.go:165 +0x88
   github.com/numtide/treefmt/v2/cmd/format.Run(0x4?, 0xc0000aeaa0, 0xc0001cbc48?, {0xc0000d5da0, 0x1, 0x3})
       github.com/numtide/treefmt/v2/cmd/format/format.go:157 +0x8e6
   github.com/numtide/treefmt/v2/cmd.runE(0xc0000ea480, 0xc0000aeaa0, 0xc0001bc008, {0xc0000d5da0, 0x1, 0x3})
       github.com/numtide/treefmt/v2/cmd/root.go:183 +0x6c6
   github.com/numtide/treefmt/v2/cmd.NewRoot.func1(0xc000180500?, {0xc0000d5da0?, 0x7?, 0x8c0546?})
       github.com/numtide/treefmt/v2/cmd/root.go:37 +0x32
   github.com/spf13/cobra.(*Command).execute(0xc0001bc008, {0xc000014110, 0x3, 0x3})
       github.com/spf13/cobra@v1.9.1/command.go:1015 +0xaaa
   github.com/spf13/cobra.(*Command).ExecuteC(0xc0001bc008)
       github.com/spf13/cobra@v1.9.1/command.go:1148 +0x46f
   github.com/spf13/cobra.(*Command).Execute(0xc000002380?)
       github.com/spf13/cobra@v1.9.1/command.go:1071 +0x13
   main.main()
       github.com/numtide/treefmt/v2/main.go:12 +0x18
   ```

**Expected behavior**

[...]

PR #595, opened 2025-05-25
fix: `treefmt .` no longer segfaults if `.` is root and symlink

jfly:
In short, the old code would massage the given paths to relative paths (in the `resolvePaths` function I removed). It operated under an impossible-to-satisfy requirement: to return relative paths that when joined with the project root would create paths to regular files/folders (*not* symlinks). This is possible for all paths except `.`: if your project root is a symlink, then `join(root, ".") == root`, which is still a symlink.

Rather than try to fix that flawed assumption, I opted to refactor things. I think that was the right call: excluding the tests I added, this is a net reduction in code!

Basically, I moved moved `resolvePath` into `walk.go`, and interleaved the remaining `resolvePaths` logic into `walk.go::NewCompositeReader`. A happy side effect of this is that we no longer have 2 functions with very similar names (`resolvePaths` and `resolvePath`) that do different things.

This fixes https://github.com/numtide/treefmt/issues/594


jfly:
@brianmcgee, could I get your eyes on this?

brianmcgee (approved):

PR #601, opened 2025-06-19
feat: add jujutsu walker

delafthi:
This PR introduces a new Jujutsu walker to numtide/treefmt, modeled after the existing Git walker but tailored to Jujutsu repositories. Key differences include:

- The list command does not update the index automatically, so untracked files must be explicitly added via `jj` commands.

This addition enables proper handling of Jujutsu ignore rules without requiring co-located Git repos, improving integration and file traversal accuracy for Jujutsu users.

I lack Go experience and primarily copied, and adapted the existing Git walker by adjusting commands. I would welcome guidance or collaboration to develop a more idiomatic and robust implementation.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1536,6 +1536,218 @@ func TestGit(t *testing.T) {
 	)
 }
 
+func TestJujutsu(t *testing.T) {
+	as := require.New(t)
+
+	test.SetenvXdgConfigDir(t)
+	tempDir := test.TempExamples(t)
+	configPath := filepath.Join(tempDir, "/treefmt.toml")
+
+	test.ChangeWorkDir(t, tempDir)
+
+	// basic config
+	cfg := &config.Config{
+		FormatterConfigs: map[string]*config.Formatter{
+			"echo": {
+				Command:  "echo", // will not generate any underlying changes in the file
+				Includes: []string{"*"},
+			},
+		},
+	}
+
+	test.WriteConfig(t, configPath, cfg)
+
+	// init a jujutsu repo
+	jjCmd := exec.Command("jj", "git", "init")
+	as.NoError(jjCmd.Run(), "failed to init jujutsu repository")
+
+	// run treefmt before adding anything to the jj index
+	// Jujutsu depends on updating the index with a `jj` command. So, until we do
+	// that, the treefmt should return nothing, since the walker is executed with
+	// `--ignore-working-copy` which does not update the index.
+	treefmt(t,
+		withConfig(configPath, cfg),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 0,
+			stats.Matched:   0,
+			stats.Formatted: 0,
+			stats.Changed:   0,
+		}),
+	)
+
+	// update jujutsu's index
+	jjCmd = exec.Command("jj")
+	as.NoError(jjCmd.Run(), "failed to update the index")
+
+	// This is our first pass, since previously the files were not in the index. This should format all files.
+	treefmt(t,
+		withConfig(configPath, cfg),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 32,
+			stats.Matched:   32,
+			stats.Formatted: 32,
+			stats.Changed:   0,
+		}),
+	)
+
+	// create a file which should be in .gitignore
+	f, err := os.CreateTemp(tempDir, "test-*.txt")
+	as.NoError(err, "failed to create temp file")
+
+	// update jujutsu's index
+	jjCmd = exec.Command("jj")
+	as.NoError(jjCmd.Run(), "failed to update the index")
+
+	t.Cleanup(func() {
+		_ = f.Close()
+	})
+
+	treefmt(t,
+		withConfig(configPath, cfg),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 32,
+			stats.Matched:   32,
+			stats.Formatted: 0,
+			stats.Changed:   0,
+		}),
+	)
+
+	// remove python directory
+	as.NoError(os.RemoveAll(filepath.Join(tempDir, "python")), "failed to remove python directory")
+
+	// update jujutsu's index
+	jjCmd = exec.Command("jj")
+	as.NoError(jjCmd.Run(), "failed to update the index")
+
+	// we should traverse and match against fewer files, but no formatting should occur as no formatting signatures
+	// are impacted
+	treefmt(t,
+		withConfig(configPath, cfg),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 29,
+			stats.Matched:   29,
+			stats.Formatted: 0,
+			stats.Changed:   0,
+		}),
+	)
+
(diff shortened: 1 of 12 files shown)


cmd/root_test.go
```diff
@@ -1536,6 +1536,232 @@ func TestGit(t *testing.T) {
...
+
+	// create a file which should be in .gitignore
+	f, err := os.CreateTemp(tempDir, "test-*.txt")
+	as.NoError(err, "failed to create temp file")
+
+	// add everything to the index (disable signing for testing), this command shoud ignore files in .gitignore
+	jjCmd = exec.Command("jj", "file", "track", "--config", "signing.backend=none", ".")
+	as.NoError(jjCmd.Run(), "failed to add everything to the index")
+
+	// update jujutsu's index (disable signing for testing)
```
jfly, line 1605: > (disable signing for testing)

This is pretty repetitive. Can we instead create a config file for the test?

cmd/root_test.go
```diff
@@ -1536,6 +1536,232 @@ func TestGit(t *testing.T) {
...
+	}
+
+	test.WriteConfig(t, configPath, cfg)
+
+	// init a jujutsu repo (disable signing for testing)
+	jjCmd := exec.Command("jj", "git", "init", "--config", "signing.backend=none")
+	as.NoError(jjCmd.Run(), "failed to init jujutsu repository")
+
+	// run before adding anything to the index
+	// we shouldn't pick up untracked files, because here the jujutsu walker differs from the git walker
```
jfly, line 1564: I have no knowledge of jj. I can't tell if this comment is telling me something fundamentally different about jj, or if it's telling me something about treefmt's jj walker implementation. Could you rephrase this to make it clear?

cmd/root_test.go
```diff
@@ -1536,6 +1536,232 @@ func TestGit(t *testing.T) {
...
+	// try with a bad path
+	treefmt(t,
+		withArgs("-C", tempDir, "haskell", "foo"),
+		withConfig(configPath, cfg),
+		withError(func(as *require.Assertions, err error) {
+			as.ErrorContains(err, "foo not found")
+		}),
+	)
+
+	// try with a path not in the git index
```
jfly, line 1716: > git index

Is this the right term with jj?

walk/jujutsu_test.go
```diff
@@ -0,0 +1,87 @@
...
+	cmd = exec.Command("jj", "--config", "signing.backend=none")
+	cmd.Dir = tempDir
+	as.NoError(cmd.Run(), "failed to update the index")
+
+	// read empty worktree
+	statz := stats.New()
+	reader, err := walk.NewJujutsuReader(tempDir, "", &statz)
+	as.NoError(err)
+
+	files := make([]*walk.File, 33)
```
jfly, line 43: nit: The number 33 looks pretty magical here. Is there some way to get rid of it? If not, could you add some explanation?

walk/walk.go
```diff
@@ -208,7 +209,10 @@ func NewReader(
 		// for now, we keep it simple and try git first, filesystem second
```
jfly, line 209: This comment is no longer accurate.

walk/jujutsu.go
```diff
@@ -0,0 +1,171 @@
...
+		})
+
+		// create a new scanner for reading the output
+		j.scanner = bufio.NewScanner(r)
+	}
+
+	nextLine := func() (string, error) {
+		line := j.scanner.Text()
+
+		if len(line) == 0 || line[0] != '"' {
```
jfly, line 66: > `line[0] != '"'`

This needs some explanation. What are we skipping? Is this really safe? Is there a more structured (json?) output we could consume instead?

jfly (commented):
The approach here looks good to me.

In the future, it could make sense to create some abstraction over `git` and `jj`, but IMO that's not warranted here. Copying and tweaking is good, and the tests you're adding are super valuable if we ever feel the need to do a future refactor.

jujutsu/jujutsu.go
```diff
@@ -0,0 +1,28 @@
...
+
+const TreeRootCmd = "jj workspace root"
+
+func IsInsideWorktree(path string) (bool, error) {
+	// check if the root is a jujutsu repository
+	cmd := exec.Command("jj", "workspace", "root")
+	cmd.Dir = path
+
+	err := cmd.Run()
+	if err != nil {
```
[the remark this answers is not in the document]
delafthi: If `err := cmd.Run()` returns an error we check the error message. The first if statement
```go
if errors.As(err, &exitErr) && strings.Contains(string(exitErr.Stderr), "There is no jj repo in \".\"") {
  return false, nil
}
```

handles the case we are not in a `jj` repo.
The following part handles every other case including if `jj` is not installed.

```go
return false, fmt.Errorf("failed to check if %s is a jujutsu repository: %w", path, err)
```

The user should still be able to use treefmt, because here we just return that the root of the repository could not be found with jujutsu. So we proceed finding the root here https://github.com/delafthi/treefmt/blob/c6d51bc7b3c35252d44a6b72fc50cf6d7f952413/config/config.go#L348.

cmd/root_test.go
delafthi: I agree, I'll change that.

cmd/root_test.go
delafthi: No, thanks for catching that

config/config.go
```diff
@@ -329,6 +330,21 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
 			}
 		}
 
+		// attempt to resolve with jujutsu
```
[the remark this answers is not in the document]
delafthi: Currently, I added `cfg.TreeRoot == ""` as a guard:

```go
if cfg.TreeRoot == "" && (cfg.Walk == walk.Auto.String() || cfg.Walk == walk.Jujutsu.String()) {
  logger.Infof("attempting to resolve tree root using jujutsu: %s", jujutsu.TreeRootCmd)
  ...
```

cmd/root_test.go
delafthi: I created a `test/config` directory containing a basic Jujutsu configuration. This config is included in the test environment by setting the `XDG_CONFIG_HOME` environment variable.

delafthi:
Additionally, I rebase unto latest main to fix the failing tests in `TestGit` (`cmd/root_test.go`).

walk/jujutsu.go
delafthi: This function returns lines without quotes. If a line is empty or does not start with a quote, it is returned immediately. Lines that are quoted are unquoted on line 71 before being returned. However, I'm not sure why this check is necessary, as neither `git ls-files` nor `jj file list` appear to return quoted paths. I have this in here because `walk/git.go` does too.

delafthi: I observed that `git ls-files` in this repository returns:

```
...
"test/examples/emoji \360\237\225\260\357\270\217/README.md"
...
```

Whereas `jj file list` outputs:

```
...
test/examples/emoji 🕰️/README.md
...
```

Therefore, it seems that unquoting can be removed for Jujutsu.

brianmcgee:
@delafthi I think what you're seeing is a brittle test. I've been seeing it periodically and trying to get to the bottom of it with now joy :disappointed: 

I'll try to find some time to review this week.

PR #610, opened 2025-07-21
chore: require that formatters exit nonzero on error

jfly:
I opted to put this in what felt like a logical order, but it does mean that other rules ("Idempotent" and "Reliable") got renumbered.

How do we feel about that? IMO, we should probably have version number for the spec.

The naming and styling of the changes here are intended to be consistent with https://github.com/numtide/treefmt/pull/609


brianmcgee (approved):

brianmcgee:
I agree, a version number for the spec is probably needed at this point.

PR #611, opened 2025-07-21
fix: includes/excludes and stdin

brianmcgee:


jfly (commented):
Thanks! Makes sense, logic is clean. Needs a test, otherwise LGTM.

PR #635, opened 2025-09-20
chore: fix flaky tests

jfly:
I found that `TestGit` (and `TestJujutsu`, which was copied from `TestGit`) were failing intermittently. This took me some time to track down: the issue is an interaction between a few things:

- This test configures a formatter to run on all files (`"*"`), which importantly includes the `treefmt.toml` file at the root of the temp test project
- This test invokes treefmt with `withConfig(configPath, cfg)`. Under the hood, `withConfig` always writes to the `treefmt.toml` config file, which (possibly) means updating the mtime of `treefmt.toml` file.
- treefmt computes cache signatures using mtimes at 1-second granularity (presumably because this the only cross filesystem compatible option).

In effect, if the test ran quickly (under 1 second), then `treefmt.toml` would get formatted exactly 1 time on the first invocation of treefmt. However, if the test ran slow enough that the mtime actually changed throughout the test, then `treefmt.toml` would get formatted multiple times.

I could have fixed this in a few ways. I opted to remove the unnecessary calls to `withConfig(configPath, cfg)` sprinkled throughout the test. Furthermore, to make this less likely to recur, I changed `WriteConfig` to ensure that the resulting file *always* has a new mtime, even if this requires setting it to the future.

### DIFF
--- a/test/test.go
+++ b/test/test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"syscall"
 	"testing"
 	"time"
 
@@ -28,6 +29,16 @@ func SetenvXdgConfigDir(t *testing.T) {
 func WriteConfig(t *testing.T, path string, cfg *config.Config) {
 	t.Helper()
 
+	oldInfo, err := os.Lstat(path)
+
+	switch {
+	case os.IsNotExist(err):
+		// It's fine if there was no old config file, the new file is guaranteed to appear new =)
+		oldInfo = nil
+	case err != nil:
+		t.Fatalf("failed to stat old config file: %v", path)
+	}
+
 	f, err := os.Create(path)
 	if err != nil {
 		t.Fatalf("failed to create a new config file: %v", err)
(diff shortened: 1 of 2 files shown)


test/test.go
```diff
@@ -37,6 +48,35 @@ func WriteConfig(t *testing.T, path string, cfg *config.Config) {
...
+	// Note: we're comparing `Unix()` (which is only 1 second granularity) for consistency with
+	// `walk.go::formatSignature`.
+	if oldInfo != nil && oldInfo.ModTime().Unix() == newInfo.ModTime().Unix() {
+		// No change to atime.
+		sysStat, ok := newInfo.Sys().(*syscall.Stat_t)
+		if !ok {
+			t.Fatalf("failed to cast stat Sys")
+		}
+
+		atime := sysStat.Atim
```
jfly, line 69: This fails on darwin:

```
treefmt> test/test.go:69:20: sysStat.Atim undefined (type *syscall.Stat_t has no field or method Atim)
```

I don't even want to know the atime, I just want to change the mtime, but `Lutimes` wants both an atime and a mtime.

Easiest fix might be to just set atime to the current timestamp, as we don't really care about atime in treefmt.

brianmcgee:
This has been haunting us for a while!

Issue #334, opened 2024-07-03
Allow providing the max batch size as a cli arg


brianmcgee:
**Is your feature request related to a problem? Please describe.**

Currently, all batching is hard coded to a maximum of `1024`. In certain circumstances, it might be beneficial to specify smaller batch sizes to improve concurrency. 

**Describe the solution you'd like**

Add a cli arg, `--batch-size` which controls the max batch size.

brianmcgee:
Yeah, worth revisiting this. I'll have some time for treefmt at the end of this week. It would be nice to get this in before releasing #571

jfly:
I chatted with @brianmcgee about this: gonna hack on this a bit, shouldn't take too long.

PR #636, opened 2025-09-20
feat: add a `max-batch-size` option for formatters

jfly:
This is part of https://github.com/numtide/treefmt/issues/334.

### DIFF
--- a/format/formatter.go
+++ b/format/formatter.go
@@ -20,6 +20,10 @@ import (
 	"mvdan.cc/sh/v3/interp"
 )
 
+const (
+	BatchSize = 1024
+)
+
 var (
 	ErrInvalidName = errors.New("formatter name must only contain alphanumeric characters, `_` or `-`")
 	// ErrCommandNotFound is returned when the Command for a Formatter is not available.
@@ -44,6 +48,14 @@ type Formatter struct {
 
 func (f *Formatter) Name() string {
 	return f.name
+}
+
+func (f *Formatter) MaxBatchSize() int {
+	if f.config.MaxBatchSize == nil {
+		return BatchSize
+	}
+
+	return *f.config.MaxBatchSize
 }
 
 func (f *Formatter) Priority() int {
@@ -78,6 +90,10 @@ func (f *Formatter) Hash(h hash.Hash) error {
 }
 
 func (f *Formatter) Apply(ctx context.Context, files []*walk.File) error {
+	if len(files) > f.MaxBatchSize() {
+		return fmt.Errorf("formatter cannot format %d files at once (max batch size: %d)", len(files), f.MaxBatchSize())
+	}
+
 	start := time.Now()
 
 	// construct args, starting with config
--- a/format/scheduler.go
+++ b/format/scheduler.go
@@ -145,8 +145,10 @@ func (s *scheduler) schedule(ctx context.Context, key batchKey, batch []*walk.Fi
 		for _, name := range key.sequence() {
 			formatter := s.formatters[name]
 
-			if err := formatter.Apply(ctx, batch); err != nil {
-				formatErrors = append(formatErrors, err)
+			for chunk := range slices.Chunk(batch, formatter.MaxBatchSize()) {
+				if err := formatter.Apply(ctx, chunk); err != nil {
+					formatErrors = append(formatErrors, err)
+				}
 			}
 		}
 
--- a/cmd/format/format.go
+++ b/cmd/format/format.go
@@ -22,10 +22,6 @@ import (
 	bolt "go.etcd.io/bbolt"
 )
 
-const (
-	BatchSize = 1024
-)
-
 var ErrFailOnChange = errors.New("unexpected changes detected, --fail-on-change is enabled")
 
 func Run(v *viper.Viper, statz *stats.Stats, cmd *cobra.Command, paths []string) error {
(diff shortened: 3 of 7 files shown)


format/formatter.go
```diff
@@ -46,6 +46,14 @@ func (f *Formatter) Name() string {
 	return f.name
 }
 
+func (f *Formatter) MaxBatchSize() int {
+	if f.config.MaxBatchSize == nil {
+		return 1024 //<<<
```
jfly, line 51: We have 2 other `BatchSize` constants in treefmt already;

```
cmd/format/format.go
26:	BatchSize = 1024

walk/walk.go
30:	BatchSize = 1024
```

Not sure if we really want a 3rd, or if we should instead import one of those here.

brianmcgee:
@jfly can you rebase this on the latest and greatest

format/formatter.go
```diff
@@ -78,6 +86,24 @@ func (f *Formatter) Hash(h hash.Hash) error {
 }
 
 func (f *Formatter) Apply(ctx context.Context, files []*walk.File) error {
+	for i := 0; i < len(files); i += f.MaxBatchSize() {
```
brianmcgee: I don't understand what this is doing and why the need to introduce the `apply` method.

format/formatter.go
jfly: Done! Well, actually I moved it from `format.go` to `formatter.go` because `format.go` already imports `formatter.go` and go (reasonably) wouldn't let me introduce a circular dependency.

format/formatter.go
```diff
@@ -78,6 +86,24 @@ func (f *Formatter) Hash(h hash.Hash) error {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+func (f *Formatter) apply(ctx context.Context, files []*walk.File) error {
+	if len(files) > f.MaxBatchSize() {
+		//<<< learn me some go >>>
```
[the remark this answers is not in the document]
jfly: How would we do that? That codepath is invoking `formatter.apply` for all formatters for that filetype, and tracking if any errors occurred. There will be possibly many errors, and we [boil that down to a single boolean](https://github.com/jfly/treefmt/blob/c24e1f1b4fc3087f4cbfef4a0ad076bd86b129f0/format/scheduler.go#L156-L157), and ultimately return [this simple, untemplated error message](https://github.com/jfly/treefmt/blob/c24e1f1b4fc3087f4cbfef4a0ad076bd86b129f0/format/scheduler.go#L217-L219).

Do we want to collect all the errors and join them? Perhaps this should be a future improvement? I just stumbled across it while working on this PR and wanted to discuss it.

format/scheduler.go
```diff
@@ -127,6 +128,7 @@ func (s *scheduler) submit(
 	s.batches[key] = append(s.batches[key], file)
 
 	// schedule the batch for processing if it's full
+	// <<< TODO: impedance mismatch? instead compute LCM (up to some max) of max-batch-size of all matched formatters? >>>
```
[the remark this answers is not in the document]
jfly: Right now, `s.batchSize` is always 1024 (it's ultimately set to that constant [here](https://github.com/jfly/treefmt/blob/c24e1f1b4fc3087f4cbfef4a0ad076bd86b129f0/cmd/format/format.go#L126)). That means that for a formatter whose `max-batch-size` is 1, we still have to collect 1024 files before we start invoking the formatter. When I wrote this, that felt a little off to me: ideally we'd start invoking the formatter as soon as we've found its `max-batch-size` number of files. However, a batch may be formatted by multiple formatters, each with its own `match-batch-size`. Consider the scenario where you're collecting files to be formatted by 2 formatters:

- `fmt5` with `max-batch-size = 5`
- `fmt3` with `max-batch-size = 3`

Imagine we have 15 files to format. If you wait to collect 5 files before scheduling the batch, you'll invoke `fmt5` 3 times with 5 files each, and `fmt3` 6 times: 3 files, 2 files, 3 files, 2 files, 3 files, 2 files. More optimal would be to collect `LCM(3, 5) = 15` files before scheduling, and we'd invoke `fmt5` 3 times with 5 files each, and `fmt3` 5 times with 3 files each.

Perhaps an even simpler example would be a single formatter whose `max-batch-size` is 1023. We'll schedule the formatter with batches of size 1024, and under the hood, slice that up into slices of size 1024 + 1 over and over.

Now with a little bit of time and space from this code, this feels kind of silly to me, but maybe not worth doing anything about (or at least not worth holding this PR up over). WDYT?

format/formatter.go
jfly: This is just chunking the supplied list of files into chunks no larger than what our formatter can handle. `apply` asserts that the supplied list of files is no larger than the formatter can handle. Happy to rename/restructure these if you have ideas.

Perhaps it would make sense for the caller of `Apply` to do the chunking, and then this code could simply assert that it's never invoked with a too-large chunk? Please see my comment in format/scheduler.go for how this implicit chunking behavior can result in suboptimal chunking overall, which may be cleaner to resolve if the chunking is the responsibility of the scheduler.

jfly (commented):
> @jfly can you rebase this on the latest and greatest

Done!

format/scheduler.go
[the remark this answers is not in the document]
jfly: Sounds good. Marking this resolved.

Issue #154, opened 2022-05-01
ignore submodules


nrdxp:
For many uses of submodules the project doesn't necessarily own the rights to the submodule directory so formatting it is just a waste of time since the code cannot be committed. Therefore, we should have an option to skip submodules, possibly by default.

zimbatm:
I'm not opposed to it but treefmt doesn't know about git or any other SCM at the moment so it would be a jump in complexity for the project.

With the current version you can explicitly ignore all the files in a folder with:
```toml
[global]
excludes = ["./path/to/submodule/**"]
```
(see https://numtide.github.io/treefmt/treefmt-configuration.html#global)

bitbloxhub:
Noticed when using `typos` that the git walker will match submodules but not their contents, because `git ls-files` lists them. If you parse the output of `git ls-files -s` you can filter out the ones that have a filemode of `160000`.

PR #661, opened 2026-02-02
fix: ignore git submodules

bitbloxhub:
Fixes https://github.com/numtide/treefmt/issues/154.

### DIFF
--- a/walk/git.go
+++ b/walk/git.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"strings"
 
 	"github.com/charmbracelet/log"
 	"github.com/numtide/treefmt/v2/git"
@@ -38,7 +39,7 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error) 
 		r, w := io.Pipe()
 
 		// create a command which will execute from the specified sub path within root
-		cmd := exec.Command("git", "ls-files", "--cached", "--others", "--exclude-standard")
+		cmd := exec.Command("git", "ls-files", "--cached", "--others", "--exclude-standard", "--stage")
 		cmd.Dir = filepath.Join(g.root, g.path)
 		cmd.Stdout = w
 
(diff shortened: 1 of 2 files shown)


walk/git.go
```diff
@@ -54,13 +55,27 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error)
...
+		var stage, file string
+		if len(lineSplit) == 1 {
+			stage, file = "", lineSplit[0]
+		} else {
+			stage, file = lineSplit[0], lineSplit[1]
+		}
+
+		// 160000 is the mode for submodules
+		if strings.HasPrefix(stage, "160000") {
+			return "", nil
```
jfly: nit: It feels weird to me to use `""` for submodules, and for errors. How do you feel about one of these alternate ideas?

1. Return a triplet `(file, skip, err)` instead. In this case: `return "", true, nil`
2. Have this function loop until it finds a file, rather than making the caller skip the value.

walk/git.go
```diff
@@ -54,13 +55,27 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error)
 	nextLine := func() (string, error) {
```
jfly: nit: We should rename this to `nextFile` as well.

jfly (commented):
Thanks for the contribution! Could you add a test?

walk/git.go
bitbloxhub: Implemented the second solution.

walk/git.go
```diff
@@ -51,19 +52,36 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error)
...
-			return line, nil
-		}
+			var stage, file string
+			if len(lineSplit) == 1 {
+				stage, file = "", lineSplit[0]
+			} else {
+				stage, file = lineSplit[0], lineSplit[1]
+			}
+
+			// 160000 is the mode for submodules
```
jfly: May a quick sentence explaining why we skip submodules? ("they're their own projects with possibly their own formatting rules")

walk/git.go
```diff
@@ -51,19 +52,36 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error)
 		g.scanner = bufio.NewScanner(r)
 	}
 
-	nextLine := func() (string, error) {
-		line := g.scanner.Text()
+	nextFile := func() (string, error) {
+		for line := g.scanner.Text(); len(line) > 0; line = g.scanner.Text() {
```
jfly: I don't totally grok why we now have both a `len(line) > 0` check here and a `len(file) > 0` check down below. Are they both necessary?

walk/git.go
```diff
@@ -51,19 +52,36 @@ func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error)
...
+			var stage, file string
+			if len(lineSplit) == 1 {
+				stage, file = "", lineSplit[0]
+			} else {
+				stage, file = lineSplit[0], lineSplit[1]
+			}
+
+			// 160000 is the mode for submodules
+			if strings.HasPrefix(stage, "160000") {
+				g.scanner.Scan()
```
jfly: I don't understand why we need this here. Won't the for loop do the right thing when we `continue`?

bitbloxhub: No, because this loop just calls `g.scanner.Text()`, and `g.scanner.Scan()` is called in the loop that we call `nextFile` in. If you want, I could refactor it so we call `Scan` in the `nextFile` function.

jfly: Ohhh, got it. Your call, whatever you prefer.

aduh95:
Reverting this fixed https://github.com/numtide/treefmt/issues/677 for us

bitbloxhub:
Unfortunately I don't see any way to `ls-files` only things present in the current worktree, so I can't see any trivial way to solve this without potentially making performance even worse. I still don't see any reason why it should be hanging, though, if it is hanging, although I'm not really a big Go developer.

PR #669, opened 2026-02-27
feat: look for config file in `.config` subdirectory

### LINKED ISSUE #668: Supprt the `.config` directory in the project root

- The project root configuration scene in general is getting a bit messy.

- Check a `.config` folder in the root of the project for the `treefmt.toml` file. For a visual showcase of what that would look like, check this repository: https://github.com/Michael-C-Buckley/nixos

- Using the environment variable to change the location of the `treefmt.toml` file.

**Additional context**

- The "config directory proposal": https://github.com/pi0/config-dir

brianmcgee:
Closes #668

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -356,6 +357,11 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
 			)
 
 			cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
+
+			// if the config file was located inside a `.config` directory, we need to go up one level
+			if path.Base(cfg.TreeRoot) == ".config" {
+				cfg.TreeRoot = filepath.Dir(cfg.TreeRoot)
+			}
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -353,6 +353,11 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
 			)
 
 			cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
+
+			// if the config file resides within a `.config` folder, we use its parent folder as the tree root
+			if strings.HasSuffix(cfg.TreeRoot, ".config") {
```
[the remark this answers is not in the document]
brianmcgee: I gave it a try, but it got twisty real fast. 

You rightly pointed out the dodgy suffix-based matching, which I've beefed up, and added a test case for `foo.config` to be sure. But this is the simplest implementation.

config/config.go
```diff
@@ -356,6 +357,11 @@ func determineTreeRoot(v *viper.Viper, cfg *Config, logger *log.Logger) error {
 			)
 
 			cfg.TreeRoot = filepath.Dir(v.ConfigFileUsed())
+
+			// if the config file was located inside a `.config` directory, we need to go up one level
+			if path.Base(cfg.TreeRoot) == ".config" {
```
jfly, line 362: Similar question as before: would this do something screwy if the project folder were named `.config`?

brianmcgee: I would view that as user error and not something we can reasonably be expected to account for.

PR #672, opened 2026-03-07
Add a path agnostic requirement to the formatter spec

jfly:
This is an alternative to implementing a stdin specification
(https://github.com/numtide/treefmt/pull/586).

### DIFF
--- a/docs/site/reference/formatter-spec.md
+++ b/docs/site/reference/formatter-spec.md
@@ -68,3 +68,8 @@ outputs.
 ### 5. Reliable
 
 We expect the formatter to be reliable and not break the semantics of the formatted files.
+
+### 6. Path agnostic
+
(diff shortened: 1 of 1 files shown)


jfly:
After more thinking, it's not just ormolu that voilates this proposed spec, [`prettier` does](https://github.com/numtide/treefmt-nix/issues/530#issuecomment-3492073920), as [does `dprint`](https://github.com/numtide/treefmt-nix/issues/530#issuecomment-5017052530).

@brianmcgee, I think the best option is for us to land [the stdin spec](https://github.com/numtide/treefmt/pull/586), or remove treefmt's (fundamentally broken) `--stdin` feature. WDYT?

jfly:
Closing this now that https://github.com/numtide/treefmt/pull/586 has landed.

PR #679, opened 2026-03-19
chore: rename `max-batch-size` to `violates-rule-1`

jfly:
`max-batch-size` feels a bit silly to me: it's something people are only ever going to set to 1 as a workaround for formatters that violate rule 1 of the formatter spec.

I say let's call a spade a spade and make this a boolean with a name that makes you feel bad for using it.

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -2502,19 +2502,19 @@ func TestConcurrentInvocation(t *testing.T) {
 	as.NoError(eg.Wait())
 }
 
-func TestViolatesRule1(t *testing.T) {
+func TestNoPositionalArgSupport(t *testing.T) {
 	tempDir := test.TempExamples(t)
 	configPath := filepath.Join(tempDir, "/treefmt.toml")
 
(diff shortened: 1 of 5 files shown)


MagicRB:
isn't there a max arglist size on Linux? if the paths to the files are long and there are 1024 of such paths, you might go over the limit no?

jfly:
@brianmcgee, I was hoping to land this before the next release so it wouldn't be a breaking change.

With 2.5.0 out, I'd suggest we either land this quickly, or abandon it. Thoughts?

brianmcgee:
My bad I'll merge and recut. I messed up the nixpkgs PR anyway

brianmcgee:
Making an executive decision and going with Matt's suggestion of `no-positional-arg-support`. I think it's clearer than `violates-rule-1` but also conveys that the formatter is deviating from the norm. I'll merge this after dinner and recut the `2.5.0` release before I fix the bad nixpkgs PR.

jfly:
Thanks!

PR #687, opened 2026-04-15
feat: refine renovate config

brianmcgee:
- group all non-major updates into a single PR
- wait at least 7 days after a dependency has been updated
- enable lock file maintenance for `flake.lock`

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -1,4 +1,7 @@
 {
     "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-    "extends": ["config:recommended"]
+    "extends": ["config:recommended"],
+    "group:allNonMajor": true,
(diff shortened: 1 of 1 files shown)


jfly (approved):
Not an expert with renovate, but this looks good to me! I'm definitely supportive of grouping updates into fewer PRs.

jfly:
> I'm definitely supportive of grouping updates into fewer PRs

For the record, I'm so supportive of this that I don't even care if the the updates are major or minor. Our tests should be telling us about breakages either way, and in my experience in other ecosystems, just because a package technically has breaking changes doesn't mean we're affected by that breaking change (for example, removing some deprecated functionality may be a breaking change, but we don't care if we weren't using that deprecated functionality).

PR #698, opened 2026-05-15
feat: add custom walker support

tobim:
This adds configurable custom walkers. A custom walker runs a configured command, reads newline- or NUL-delimited paths from stdout, normalizes them relative to treefmt's tree root, and feeds existing regular files into the formatter pipeline.

This PR is stacked on #705, which adds the shared path stream reader used by custom walkers.

Example: Format a git repo + a specific submodule:

```toml
walk = "repoPlusSubmodule"

[walker.repoPlusSubmodule]
command = "bash"
options = [
  "-c",
  '''
    submodule="path/to/submodule"
    {
      git ls-files --cached --others --exclude-standard --full-name
      git -C "$submodule" ls-files --cached --others --exclude-standard --full-name \
        | sed "s#^#$submodule/#"
    }
  ''',
]
```

Example: Format a pijul repo:

```toml
walk = "pijul"

[walker.pijul]
command = "pijul"
options = ["ls"]
```

### DIFF
--- a/cmd/init/init.toml
+++ b/cmd/init/init.toml
@@ -48,9 +48,17 @@
 # verbose = 2
 
 # The method used to traverse the files within the tree root
-# Currently, we support 'auto', 'git', 'jujutsu', or 'filesystem'
+# Built-in values are 'auto', 'git', 'jujutsu', and 'filesystem'
+# You can also set this to the name of a configured custom walker
 # Env $TREEFMT_WALK
 # walk = "filesystem"
+# walk = "mywalker"
+
+# Custom walkers are configured with [walker.<name>]
+# They receive requested file and directory paths as positional args
+# [walker.mywalker]
+# command = "command-to-run"
+# options = []
 
 [formatter.mylanguage]
 # Command to execute
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -1814,6 +1814,65 @@ func TestJujutsu(t *testing.T) {
 	)
 }
 
+func TestCustomWalker(t *testing.T) {
+	tempDir := test.TempExamples(t)
+	configPath := filepath.Join(tempDir, "/treefmt.toml")
+
+	test.ChangeWorkDir(t, tempDir)
+
+	cfg := &config.Config{
+		Walk: "mywalker",
+		WalkerConfigs: map[string]*config.Walker{
+			"mywalker": {
+				Command: "bash",
+				Options: []string{
+					"-c",
+					`if [ "$#" -eq 0 ]; then set -- go haskell; fi
+for path in "$@"; do
+	case "$path" in
+	go) printf '%s\n' go/main.go go/go.mod ;;
+	haskell) printf '%s\n' haskell/Foo.hs ;;
+	*) printf '%s\n' "$path" ;;
+	esac
+done`,
+					"custom-walker",
+				},
+			},
+		},
+		FormatterConfigs: map[string]*config.Formatter{
+			"echo": {
+				Command:  "echo", // will not generate any underlying changes in the file
+				Includes: []string{"*"},
+			},
+		},
+	}
+
+	test.WriteConfig(t, configPath, cfg)
+
+	treefmt(t,
+		withConfig(configPath, cfg),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 3,
+			stats.Matched:   3,
+			stats.Formatted: 3,
+			stats.Changed:   0,
+		}),
+	)
+
+	treefmt(t,
+		withArgs("--clear-cache", "go"),
+		withConfig(configPath, cfg),
+		withNoError(t),
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 2,
+			stats.Matched:   2,
+			stats.Formatted: 2,
+			stats.Changed:   0,
+		}),
+	)
+}
+
 func TestTreeRootCmd(t *testing.T) {
 	as := require.New(t)
 
--- a/config/config.go
+++ b/config/config.go
@@ -45,6 +45,7 @@ type Config struct {
 	Stdin                 bool     `mapstructure:"stdin"                   toml:"-"` // not allowed in config
 
 	FormatterConfigs map[string]*Formatter `mapstructure:"formatter" toml:"formatter,omitempty"`
+	WalkerConfigs    map[string]*Walker    `mapstructure:"walker"    toml:"walker,omitempty"`
 
 	Global struct {
 		// Deprecated: Use Excludes
@@ -66,6 +67,13 @@ type Formatter struct {
 	// Does this formatter violate [rule 1] of the formatter spec?
 	// [rule 1]: https://treefmt.com/latest/reference/formatter-spec/#1-files-passed-as-arguments
 	NoPositionalArgSupport *bool `mapstructure:"no-positional-arg-support" toml:"no-positional-arg-support"`
+}
+
+type Walker struct {
+	// Command is the command to invoke when walking the tree.
+	Command string `mapstructure:"command" toml:"command"`
+	// Options are an optional list of args to be passed to Command.
+	Options []string `mapstructure:"options,omitempty" toml:"options,omitempty"`
 }
 
 // SetFlags appends our flags to the provided flag set.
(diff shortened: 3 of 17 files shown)


jfly (commented):
@tobim, thanks for the contribution! I am supportive of this change, but haven't had a chance to read the code yet.

I wanted to warn you that https://github.com/numtide/treefmt/pull/694 touches some of the same code you've touched here. It may land before this PR, apologies if the conflicts are annoying to deal with.

docs/site/getting-started/configure.md
```diff
@@ -433,6 +442,35 @@ Run as if `treefmt` was started in the specified working directory instead of th
...
+### `command`
+
+The command to invoke when walking the tree.
+`treefmt` runs the command from the tree root.
+The command must write one path per line to `stdout`.
+Each path must be relative to the tree root.
+
+When you pass directory paths to `treefmt`, the walker command still runs for the tree root.
+`treefmt` filters the command output to the requested directories.
+The walker command doesn't need to implement path argument handling.
```
jfly: While I appreciate the simplicity of this decision, I'm not sure it's the right decision. I imagine people are most inclined to use the "format a directory" feature in very large repos, where the time spent for some VCS to determine which files are in the entire repo might be a lot slower than discovering just the files in a specific directory. https://github.com/numtide/treefmt/pull/694 is an example of someone reporting just how slow it is for git to report all files in a repo.

How crazy would it be for the walk command to have to implement path argument handling? IIUC, it would be straightforward for `git`.

cmd/init/init.toml
```diff
@@ -48,9 +48,16 @@
 # verbose = 2
 
 # The method used to traverse the files within the tree root
-# Currently, we support 'auto', 'git', 'jujutsu', or 'filesystem'
+# Built-in values are 'auto', 'git', 'jujutsu', and 'filesystem'
+# You can also set this to the name of a configured custom walker
 # Env $TREEFMT_WALK
 # walk = "filesystem"
+# walk = "myWalker"
```
jfly: I wonder if we should instead do `walker.myWalker` (or perhaps `custom-walker.myWalker`) here. Then we wouldn't have to worry about (or detect) conflicts with the builtin walkers.

config/config.go
```diff
@@ -233,6 +245,47 @@ func FromViper(v *viper.Viper) (*Config, error) {
...
+
+		if _, ok := cfg.WalkerConfigs[cfg.Walk]; !ok {
+			for name, walkerCfg := range cfg.WalkerConfigs {
+				if strings.EqualFold(name, cfg.Walk) {
+					cfg.WalkerConfigs[cfg.Walk] = walkerCfg
+					ok = true
+
+					break
+				}
+			}
```
jfly: Is this really worth it to be case insensitive? Are we case insensitive for the builtin walkers?

config/config.go
```diff
@@ -200,6 +208,10 @@ func FromViper(v *viper.Viper) (*Config, error) {
 		return nil, fmt.Errorf("failed to unmarshal config: %w", err)
 	}
 
+	if cfg.Walk == "" {
+		cfg.Walk = walk.Auto.String()
+	}
```
jfly: Why do we need this? Don't we default to auto [here](https://github.com/numtide/treefmt/blob/f0fb3071b90832561815fdd5565bff6979d06dbb/config/config.go#L142)?

walk/custom.go
```diff
@@ -0,0 +1,214 @@
...
+	return n, err
+}
+
+func (c *CustomReader) file(line string) (*File, bool, error) {
+	line = strings.TrimSuffix(line, "\r")
+	if line == "" {
+		return nil, false, nil
+	}
+
+	relPath, err := c.relativePath(line)
```
jfly: I believe the docs you wrote said that the custom walker must emit relative paths. Reading this code, it looks like it's OK if it emits absolute paths (as we'll convert them to relative here). Am I reading this correctly? Want to update the docs accordingly?

walk/walk.go
```diff
@@ -210,29 +233,33 @@ func NewReader(
 		reader Reader
 	)
 
-	switch walkType {
-	case Auto:
-		// for now, we keep it simple and try git first, jujutsu second, and filesystem last
-		reader, err = NewReader(Git, root, path, db, statz)
-		if err != nil {
-			reader, err = NewReader(Jujutsu, root, path, db, statz)
+	if selector.Custom != nil {
```
jfly: This diff would be a lot simpler, and the `if selector.Custom == nil && selector.Type == Stdin` change below wouldn't have had to happen if we introduced a new `selector.Type == Custom`. Did you consider that?

cmd/root_test.go
```diff
@@ -1814,6 +1814,67 @@ func TestJujutsu(t *testing.T) {
...
+		withStats(t, map[stats.Type]int{
+			stats.Traversed: 3,
+			stats.Matched:   3,
+			stats.Formatted: 3,
+			stats.Changed:   0,
+		}),
+	)
+
+	treefmt(t,
+		withArgs("-c", "go"),
```
jfly: Please use the long form in tests:

```
		withArgs("--clear-cache", "go"),
```

jfly (commented):
Thanks! This looks like a solid approach.

Not a blocker, but I'm not in love with a 3rd copy of the walk code. I didn't read it particularly closely. I think 2 copies (jj and git) is OK. 3 is when it probably makes sense to to think of an abstraction. Did you think about that at all?

> Accept documented camel-case walker names even when Viper normalizes table keys

What is `camel-case`? That's `kebab-case`. This is `camelCase` (or `CamelCase`).

> Fix custom walker pipe handling

I did not have time to grok this commit during my review. I'll need some more handholding before I can approve it. Does whatever bug you're fixing here apply to the git and jj walkers as well? Is it related to this bug @mic92 found in https://github.com/numtide/treefmt/pull/694 (commit labeled "walk: unblock producers when Close() is called before EOF").

docs/site/getting-started/configure.md
tobim: I admit that I didn't pay too much attention to optimizing for subtrees. It's no concern for my own use cases.

For a regular git walker this might not be a problem, but as soon as you want a more custom file list - like in the first example from the PR description - implementing the filtering yourself becomes error-prone rather quickly.

Maybe we'll add a toggle so the user can switch between both methods?

tobim:
Thank you for the review. I hope to be able to address the individual comments in the next few days.

> Not a blocker, but I'm not in love with a 3rd copy of the walk code. I didn't read it particularly closely. I think 2 copies (jj and git) is OK. 3 is when it probably makes sense to to think of an abstraction. Did you think about that at all?

I didn't really consider that because I wanted to keep the new path somewhat isolated.
We could also replace `git` and `jj` with built in custom walkers now that the machinery is there.

> Does whatever bug you're fixing here apply to the git and jj walkers as well? Is it related to this bug @Mic92 found in https://github.com/numtide/treefmt/pull/694 (commit labeled "walk: unblock producers when Close() is called before EOF").

I'll have to look into that.

jfly:
> I didn't really consider that because I wanted to keep the new path somewhat isolated.
> We could also replace git and jj with built in custom walkers now that the machinery is there.

This sounds reasonable to me

walk/custom.go
tobim: Updated the docs.

config/config.go
tobim: removed

tobim:
So I did somewhat of a refactor of the walker component now. Now there is a generic `PathStreamReader` abstraction that consumes a stream of paths that are separated by either a newline or a NULL character. `jj` and `git` are now constructed in terms of that `PathStreamReader`.
After that are a couple of clean-up commits that simplify the state representation and related control flow.

jfly:
> So I did somewhat of a refactor of the walker component now. Now there is a generic PathStreamReader abstraction that consumes a stream of paths that are separated by either a newline or a NULL character. jj and git are now constructed in terms of that PathStreamReader.

I am excited to read this. I'll be travelling for the next ~1 week, though, and don't expect to be able to take a look until I'm back. Could you rework this change as separate PR we could land before we add the custom walker support?

brianmcgee:
Been on holiday myself this past week or so. Just getting up to speed with this. I'm liking the direction, need to give it some proper brain time though, as it's a core change.

tobim:
I opened https://github.com/numtide/treefmt/pull/705 to split of the refactoring. This PR should now be stacked on top of that, but GitHub is crappy and you can't do that.

PR #701, opened 2026-05-27
fix: correctly detect if a path is inside the project root or not

### LINKED ISSUE #700: Files starting with `..` treated as outside tree root
**Describe the bug**

Files whose name starts with `..` (2 dots) seem to be incorrectly treated as outside of the tree root.

**To Reproduce**

```console
$ cd $(mktemp -d)
$ touch treefmt.toml foo.json ..bar.json
$ ls -A
..bar.json      foo.json        treefmt.toml
$ nix run nixpkgs#treefmt -- foo.json
traversed 1 files
emitted 0 files for processing
formatted 0 files (0 changed) in 67ms
$ nix run nixpkgs#treefmt -- ..bar.json
Error: failed to create walker: path ..bar.json not inside the tree root /private/var/folders/6z/jh9wgqld10gc2gwtt3gxy3180000gn/T/tmp.AeTNWzKKsV (relative path: ..bar.json)
```

**Expected behavior**

I expected `foo.json` and `..bar.json` to be formatted equally.

**System information**

treefmt v2.5.0 on macOS.

jfly:
We had a handful of bugs here:

1. It's not enough to check if a relative path starts with `..`, as it's valid for files to literally start with 2 dots. You need to check for the path separator as well.
2. Stdin mode incorrectly assumed that any paths that start with `..` are outside of the tree root. This is wrong in 2 ways: - If we're in a subdirectory of the project, `../foo` is still very much inside the project. - Path's can still escape without starting with `../`. For example: `foo/../../bar` escapes the project root.
3. Stdin mode incorrectly treated relative paths given by the user as relative to the project root. Now they are correctly treated as relative the current directory, and massaged into paths that are relative to the project root. The simplest approach to fixing this was to DRY up the path processing logic between stdin/non stdin mode.

This fixes https://github.com/numtide/treefmt/issues/700

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -2205,6 +2205,36 @@ func TestStdin(t *testing.T) {
 		}),
 	)
 
+	// Try a file that's more subtly outside of the project root.
+	os.Stdin = test.TempFile(t, "", "stdin", &contents)
+
+	treefmt(t,
+		withArgs("--stdin", "foo/../../test.nix"),
+		withError(func(as *require.Assertions, err error) {
+			as.ErrorContains(err, "path foo/../../test.nix not inside the tree root "+tempDir)
+		}),
+		withStderr(func(out []byte) {
+			as.Contains(string(out), "Error: failed to create walker: path foo/../../test.nix not inside the tree root")
+		}),
+	)
+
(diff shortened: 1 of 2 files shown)


brianmcgee (approved):
LGTM

PR #703, opened 2026-06-02
feat: upgrade charmbracelet/log to v2

### LINKED ISSUE #697: Upgrade to charmbracelet/log v2 to fix 10s delay under lefthook/pre-commit hooks
Related to #524, but a different trigger.

#524 fixed the case where `TERM=dumb` caused a 10s hang. My `TERM` is normal (`xterm-ghostty`), but I still hit a 10s hang when treefmt runs from a git hook via lefthook.

## What happens

Direct call: 23 ms. Same files via lefthook: 10 s. The 10s is reported by treefmt itself before any formatter runs.

```
$ time treefmt some.nix
formatted 0 files (0 changed) in 12ms        → 0.023s

$ time lefthook run pre-commit
formatted 0 files (0 changed) in 10.024s     → 10.044s
```

## Why

Treefmt sends terminal color queries (OSC 10 / 11) on startup. Lefthook gives the child process a pty, so `isTTY()` returns true and the queries get sent — but there's no real terminal on the other end, so nothing answers. Treefmt waits the full OSC timeout (~10s).

#524's fix only short-circuits when `TERM=dumb`. With a normal `TERM`, we still run the probe and still hit the timeout.

## Repro

`lefthook.yml`:
```yaml
pre-commit:
  commands:
    format:
      glob: '*.nix'
      run: treefmt {staged_files}
```

`treefmt.toml`:
```toml
[formatter.nix]
command = 'nixfmt'
includes = ['*.nix']
```

Stage a `.nix` file, run `lefthook run pre-commit`.

## Workarounds

- `TERM=dumb` in the hook env → 20 ms. Works.
- `CI=1` → also works, but changes behaviour in other tools.
- `NO_COLOR=1` → does **not** help. The probe still runs.

## Versions

- treefmt v2.5.0
- lefthook v2.1.5
- NixOS, Linux 6.18

brianmcgee:
Straightforward upgrade, just import path changes.

V2 has switched to [charmbracelet/colorprofile](https://github.com/charmbracelet/colorprofile) which doesn't send OSC queries and has better terminal color support in general.

Fixes #697


jfly (approved):
Lgtm!

Teach me about the go ecosystem: Dependabot didn't send us this update, because go lib authors tend to treat breaking changes as a new library?

How do we check if we're using other effectively unmaintained dependencies?

brianmcgee:
This was a special case. Renovate didn't pick it up because they changed the module path from `github.com/charmbracelet/log` to `charm.land/log/v2`.

PR #705, opened 2026-06-07
Add command-backed path stream walkers

### LINKED ISSUE #704: `treefmt-tree-root` config option isn't respected in Jujutsu workspaces
**Describe the bug**

When `tree-root-file` is configured in `treefmt.toml`, the jujutsu walker does not filter its file listing to the resolved tree root. It enumerates all tracked files from the jj repo root, then passes ../-prefixed relative paths to formatter glob matchers. This causes formatters to match and process files outside the intended tree root.

The git walker correctly scopes file enumeration to the tree root directory. The jujutsu walker should behave the same way.

This is only a problem in non-colocated jj repos, and jj worktrees, which also don't have a `.git`

**To Reproduce**

1. Have a jujutsu repo with this structure:
  ```shell
  repo-root/          ← .jj lives here, no .git
  ├── other-project/
  │   └── shell.nix
  └── my-project/
      └── treefmt.toml
  ```
2. my-project/treefmt.toml contains:
  ```toml
  tree-root-file = "treefmt.toml"

  [formatter.nix]
  command = "nixpkgs-fmt"
  includes = ["*.nix"]
  ```
  3. Run `treefmt --no-cache -vv` from `my-project/`.
  4. Observe that the jujutsu walker processes files from other-project/ and the entire repo:
  ```log
  INFO config: tree root: /path/to/repo-root/my-project
  DEBU walk | jujutsu: processing file: /path/to/repo-root/other-project/shell.nix
  DEBU formatter | nix: match: /path/to/repo-root/other-project/shell.nix
  ```
  5. By contrast, if the repo is colocated (has a .git directory), treefmt selects the git walker and correctly processes only files within my-project/.

**Expected behavior**

The jujutsu walker should filter its file listing to only include files within the resolved tree root, matching the behavior of the git walker. Files outside the tree root should not be walked, matched, or formatted.

**System information**

  - treefmt v2.4.0
  - jj 0.40 (non-colocated workspaces, i.e. jj-only with no .git directory)
  - macOS Darwin 25.5.0˚

**Additional context**

This particularly affects jujutsu worktrees created from a colocated repo. The main worktree has both .jj and .git, so treefmt picks the git walker (which works correctly). Additional jj worktrees only have .jj, triggering the jujutsu walker and the broken behavior.

Workaround: set TREEFMT_WALK=filesystem and use TREEFMT_TREE_ROOT env var (after removing tree-root-file from the config to avoid the "at most one of tree-root, tree-root-cmd or tree-root-file" conflict). This requires adding explicit excludes for directories like target/ that gitignore would normally handle.

tobim:
This adds a command-backed path stream reader interface for walkers that emit paths on stdout. The reader accepts newline- or NUL-delimited records, normalizes them relative to the tree root, filters requested path prefixes, and only returns existing regular files to the formatter pipeline.

The Git and Jujutsu walkers now use this shared reader. Requested file and directory paths are passed to their commands as positional filters, which lets the underlying walker prune output for subtree formatting.

The Jujutsu walker now also passes `.` when no explicit path filter is provided. This keeps `jj file list` scoped to treefmt's resolved tree root in non-colocated Jujutsu workspaces.

This also lets the filesystem walker handle multiple path filters directly and documents that paths containing newlines are unsupported.

Fixes #704.

### DIFF
--- a/walk/git.go
+++ b/walk/git.go
@@ -3,7 +3,6 @@ package walk
 import (
 	"fmt"
 
-	"charm.land/log/v2"
 	"github.com/numtide/treefmt/v2/git"
 	"github.com/numtide/treefmt/v2/stats"
 )
--- a/walk/git.go
+++ b/walk/git.go
@@ -1,143 +1,18 @@
 package walk
 
 import (
-	"bufio"
-	"context"
 	"fmt"
-	"io"
-	"os"
-	"os/exec"
-	"path/filepath"
-	"strconv"
-	"strings"
 
 	"charm.land/log/v2"
 	"github.com/numtide/treefmt/v2/git"
 	"github.com/numtide/treefmt/v2/stats"
-	"golang.org/x/sync/errgroup"
 )
 
-type GitReader struct {
-	root string
-	path string
-
-	log   *log.Logger
-	stats *stats.Stats
-
-	eg      *errgroup.Group
-	scanner *bufio.Scanner
-}
-
-func (g *GitReader) Read(ctx context.Context, files []*File) (n int, err error) {
-	// ensure we record how many files we traversed
-	defer func() {
-		g.stats.Add(stats.Traversed, n)
-	}()
-
-	nextFile := func() (string, error) {
-		for line := g.scanner.Text(); len(line) > 0; line = g.scanner.Text() {
-			lineSplit := strings.Split(line, "\t")
-
-			var stage, file string
-			// Untracked files just show as `<filename>`, while tracked files show as `<mode> <object> <stage><TAB><file>`
-			if len(lineSplit) == 1 {
-				stage, file = "", lineSplit[0]
-			} else {
-				stage, file = lineSplit[0], lineSplit[1]
-			}
-
-			// 160000 is the mode for submodules, skip them because they are separate projects that may have their own
-			// formatting rules
-			if strings.HasPrefix(stage, "160000") {
-				g.scanner.Scan()
-
-				continue
-			}
-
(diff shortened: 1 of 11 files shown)


walk/git.go
```diff
@@ -150,34 +25,10 @@ func NewGitReader(
...
-	cmd.Dir = filepath.Join(root, path)
-	cmd.Stdout = w
-
-	// execute the command in the background
-	eg.Go(func() error {
-		return w.CloseWithError(cmd.Run())
+	return NewPathStreamReader(root, statz, PathStreamConfig{
+		Name:        "git",
+		Command:     "git",
+		Options:     []string{"ls-files", "-z", "--cached", "--others", "--exclude-standard", "--full-name", "--"},
```
tobim, line 30: Removing `--stage` here should not pose a problem.
The original reason to pass that option was so submodule paths are omitted, but the `PathStreamReader` filters out directories by itself. This also speeds up the execution time of `git ls-files` quite a bit.

tobim: I would also like to remove `--others`, because those files can't be recovered when they get formatted but the user didn't want that. Imho the "git" walker should not list untracked files.
I didn't do it tho because that would make this PR a behavioral change.

tobim:
@brianmcgee anything I can do to help move this forward?

brianmcgee:
I've been tied up these past few weeks. I'll give this a review today.

brianmcgee (approved):
Solid PR. I like how incremental you made the commits, made it easy to review. 

Thanks for the contribution 🙏

PR #720, opened 2026-08-13
fix: don't append `options` and `stdin-options`

### LINKED ISSUE #718: `stdin-options` ignored if `options` are defined when running with `--stdin`
**Describe the bug**

When running treefmt with `--stdin`, I'd expect it to always consider `stdin-options` when it's specified, but unexpectedly it's ignored when `options` is specified too. This looks like a bug, as I'd like to be able to use a formatter with either `--stdin` or with normal mode without needing to change my config.

**To Reproduce**

1. Create .treefmt.toml with following content:
	```
	[formatter.c]
	includes = ["*.c"]
	command = "bash"
	options = ["-c", "echo foo > myfile.c"]
	stdin-options = ["-c", "echo bar > $path"]
	```
2. Create test file with `echo "hello" > myfile.c`
3. Run treefmt with `cat myfile.c | treefmt --stdin myfile.c`
4. Run `cat myfile.c`

**Expected behavior**

`cat myfile.c` should print "bar", instead it prints "foo".

After removing or commenting out the `options` line, repeating the steps, it correctly prints "bar".

**System information**

```
$ treefmt --version
treefmt v0499df6
```

**Additional context**

similar issue #596

jfly:
For maximal flexibility, they are supposed to be completely independent. Unfortunately, the test I added left `options` as `[]`, which meant we didn't notice this bug. I've expanded the test accordingly.

This fixes https://github.com/numtide/treefmt/issues/718.

### DIFF
--- a/format/formatter.go
+++ b/format/formatter.go
@@ -99,9 +99,6 @@ func (f *Formatter) Apply(ctx context.Context, files []*walk.File) error {
 
 	start := time.Now()
 
-	// construct args, starting with config
-	args := f.config.Options
-
 	// exit early if nothing to process
 	if len(files) == 0 {
 		return nil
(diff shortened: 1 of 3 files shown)


rszyma (approved):
tested manually and works as expected now

